PR #13, opened 2018-07-03
still needs __init__.py file to solve import problem

wyixin:


manateelazycat:
@wyixin I have merge your patches, thank you very much, good job!

PR #33, opened 2018-07-30
(fix) fix a url opening incompatibility problem with helm-mode under spacemacs

XiaoGeMa:
auto adding current directory in function eaf-open incompatibility with helm-mode under spacemacs

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -81,6 +81,7 @@
 
 ;;; Require
 (require 'dbus)
+(require 'subr-x)
 
 ;;; Code:
 (defcustom eaf-mode-hook '()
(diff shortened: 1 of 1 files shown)


manateelazycat:
@XiaoGeMa  Thanks for your patch.

But can you re-send PR? Not include the lines that just indent/blank change?
This patch is not clear that i don't know which lines is you changed.

Thanks!

 -- Andy

manateelazycat:
@XiaoGeMa I found you just add new function:

```
(defun eaf-open-url(url)
  (interactive "MOpen with EAF: ")
  (eaf-open url))
```

I don't think it's right way, it's a bug of Spacemacs that Spacemacs can't handle *interactive* function's F input type, we shouldn't not just add M with *interactive* function, it's not right way.

I can't merge this patch, thanks any way.

PR #37, opened 2018-08-06
Making the Python interpreter customizable

alastairdb:
Can we make the name of the Python command customizable? Then I can create a virtual environment with eaf's dependencies and put the full path of the environment's Python command in this variable


manateelazycat:
@alastairdb Thanks for your patch!

PR #48, opened 2018-10-05
What: open qutebrowser in eaf

mgcyung:


manateelazycat:
@mgcyung Man, it's so cool, thanks!

manateelazycat:
@mgcyung 
About send key event to qutebrowser, you should figure out which widget in qutebrowser accept key event.

Then add get_key_event_widgets function in buffer.py like this: https://github.com/manateelazycat/emacs-application-framework/blob/5baa13b5d6e21454a87d8644c49075324ad147b9/core/browser_buffer.py#L38

PR #69, opened 2019-09-17
Create eaf-open-demo, eaf-open-camera, eaf-open-qutebrowser functions

MatthewZMD:
Take out:
```
(cond ((string-equal url "eaf-demo")
           (setq app-name "demo"))
          ((string-equal url "eaf-camera")
           (setq app-name "camera"))
          ((string-equal url "eaf-qutebrowser")
           (setq app-name "qutebrowser"))
```
Because

- eaf-open is interactive "F. It is impossible to pass url as "eaf-demo" or "eaf-camera" or "eaf-qutebrowser" interactively when using counsel. It only accepts a valid file as input.

- To stay consistent with the function eaf-open-browser. It makes sense to use special functions for opening a browser and camera instead of opening files. Normal file interactions stay with eaf-open function itself.

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -24,10 +24,10 @@ Python is a perfect language to develop Qt program and it can call pretty much i
 
 ## Let me run hello word
 ```
-M-x eaf-open
+M-x eaf-open-demo
 ```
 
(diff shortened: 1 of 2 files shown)


manateelazycat:
Good point, thanks!

PR #70, opened 2019-09-17
Better EAF buffer names

MatthewZMD:
Name the buffer the actual file name instead of a truncated file path, a truncated file path most of the times will cut off parts of the file and makes the buffer name unreadable


manateelazycat:
Thanks a lot

PR #78, opened 2019-10-27
fix: if display two page, don't work on first page.

luhuaei:
Base on page index, if window display two page, translate function only  working on the last page. Therefore use click position determine page offset.

### DIFF
--- a/app/pdf-viewer/buffer.py
+++ b/app/pdf-viewer/buffer.py
@@ -422,15 +422,17 @@ class PdfViewerWidget(QWidget):
 
                 # computer absolute coordinate of page
                 x = int((pos.x() - render_x) * 1.0 / self.scale)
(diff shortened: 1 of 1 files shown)


luhuaei:
When using pdf-viewer, if the entire emacs is cut into two upper and lower windows, the pdf-viewer will not be able to scroll down with the mouse (a blank area appears, you can see the picture below), but you can scroll up and down. 

GIFrecord_2019-10-27_222248

I don't know the reason for this, but  [bd2fd46](https://github.com/manateelazycat/emacs-application-framework/pull/78/commits/bd2fd465dd2d05d10ec6ba634107a8479f3cc37e) solved the above problem. emmmm

manateelazycat:
Cool man, very good work!

PR #81, opened 2019-11-20
Function name eaf-file-open-in-dired is to vague, rename function

MatthewZMD:


manateelazycat:
Thanks for your work, but i don't like rename patch, eaf-file-open-in-dired haven't spell problem, just style rename will break user's config.

I don't like behavior that change api to break user's config.

I can't apply this patch, sorry.

MatthewZMD:
哈哈 没关系，主要是因为这名字当时也是我取的，现在觉得好难听就想改了，强迫症  :p

PR #84, opened 2019-11-23
Fix error progn: Symbol’s function definition is void: incf

MatthewZMD:


manateelazycat:
(setq buffer-read-only t) is not right, please check https://github.com/manateelazycat/emacs-application-framework/pull/83

manateelazycat:
Thanks for patch, man.

PR #85, opened 2019-11-27
Minor fixes, use const EAF_DBUS_NAME whenever possible

MatthewZMD:


manateelazycat:
Thanks for patch, man! ;)

Issue #82, opened 2019-11-21
let `eaf-open-camera` support camera taking photograph


stardiviner:
I have not found any keybinding actions on `eaf-open-camera`. Can it support taking photograph and save to file?

And if it can support Emacs' built-in major-mode keybindings `[C-h m]` to list out current major-mode keybindings/commands will be helpful. I don't know whether this is possible for eaf buffer.

MatthewZMD:
> And if it can support Emacs' built-in major-mode keybindings [C-h m] to list out current major-mode keybindings/commands will be helpful. I don't know whether this is possible for eaf buffer.

I am also looking into this

manateelazycat:
EAF's keybinding use eaf-monitor-key-event function to capture key event, not major-mode keybinding implement.

So C-h m is not easy

MatthewZMD:
我觉得如果能把它通过某种方式跟major mode绑定在一起的话会更方便用户调整配置，有什么idea吗

manateelazycat:
> 我觉得如果能把它通过某种方式跟major mode绑定在一起的话会更方便用户调整配置，有什么idea吗

现在已经是自定义按键的方式了，但是EAF必须接管输入事件转发，要不是会出现很多焦点的问题。

我觉得最好的方法就是，EAF本身Elisp这端可以做一个单独的按键Popup的功能，不用依赖 C-h m

stardiviner:
@manateelazycat 类似magit的popup按键还是挺好用的。我赞成。

manateelazycat:
> @manateelazycat 类似magit的popup按键还是挺好用的。我赞成。

欢迎PR，不能我一个人吭哧吭哧写代码，欢迎贡献。;)

stardiviner:
Ok, I used to use Magit's utility `transient` to write this popup keybinding for [bilibili.el](https://github.com/0xDEATHCODE/emms-bilibili)

manateelazycat:
C-h m fixed by https://github.com/manateelazycat/emacs-application-framework/pull/96

PR #86, opened 2019-11-27
New API, eaf-setq on the Emacs side, custom save camera directory

MatthewZMD:
`(eaf-setq sym val)` allows a brand new way to customize variables from the Emacs side that sends the variable with its values to the Python side whenever EAF starts. 
Now, `eaf-setq` gives the interface for users on the Emacs side to customize Emacs lisp that can affect Python directly, this opens a new door to EAF.

I've created an illustration of how it could work with the save camera photo feature you added few days ago. The user simply needs to
```
(eaf-setq 'eaf-camera-save-path "/new/path/")
```
in their Emacs `.emacs` config to customize the directory to save EAF Camera photos!

We now can python behaviour without actually modifying Python code!

With this capability, this also _potentially_ gives the ability for #82 , to modify EAF keybindings within Emacs Lisp, without touching the Python side.

### DIFF
--- a/app/camera/buffer.py
+++ b/app/camera/buffer.py
@@ -46,7 +46,7 @@ class AppBuffer(Buffer):
         self.buffer_widget.camera.start()
 
     def take_photo(self):
-        self.buffer_widget.take_photo()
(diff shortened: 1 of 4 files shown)


manateelazycat:
Awesome patch, thanks man!

PR #105, opened 2019-12-06
Map app names to key lists in `eaf-app-binding-alist'

clemera:
This makes it easier to add new apps, one needs to add the app name and the key variable to `eaf-app-binding-alist`, afterwards one just need to create an `eaf-open-<appname>` command.

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -265,6 +265,18 @@ Try not to modify this alist directly. Use `eaf-setq' to modify instead."
   :type 'cons
   :group 'eaf)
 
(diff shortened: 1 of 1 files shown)


MatthewZMD:
Some great refactors!

manateelazycat:
thanks man

clemera:
General question: What about backward incompatible changes? Since this isn't released on a public package repository are you ok with possibly breaking user config?

MatthewZMD:
We could start doing releases and provide CHANGELOG

Issue #110, opened 2019-12-07
Commands Like M-: and M-! get prompted twice while in any EAF program


MatthewZMD:
I noticed a weird bug, while in any EAF program's buffer and you try `M-:` or `M-!` invoking `eval-expression` or `shell-command`, it gets evaluated but the prompt appeared again right after. 

I don't have too much time this weekend to debug it, posting it here for now.

MatthewZMD:
Interactive functions like `eaf-open` also behave this way.

manateelazycat:
https://github.com/manateelazycat/snails/commit/44f77cc35f75e623a6e2a419629a16ec9b9c3fbf related bug, I believe the problem is recent eaf-monitor-key-event patch introduced.

manateelazycat:
我裸看代码的哈，没有测试，下面是我的推论：

1. eaf-monitor-key-event 这个事件过滤函数，在 https://github.com/manateelazycat/emacs-application-framework/blob/901f2e317eebc92424acadedf87feb04648bd4dd/eaf.el#L558 这里的逻辑，当按键不是EAF插件的快捷键的时候，执行一次这个按键对应的命令。

2. https://github.com/manateelazycat/emacs-application-framework/blob/901f2e317eebc92424acadedf87feb04648bd4dd/eaf.el#L569 eaf-monitor-key-event 每次执行完成后，会重新把 eaf-monitor-key-event 这个事件函数勾在 pre-command-hook 上，避免按键出异常后 eaf-monitor-key-event 这个函数无法执行的问题

3. EAF的 eaf-monotor-key-event 这个函数每一个事件循环只能执行一次，如果执行两次就会导致按键被调用了两次，就像我最近的修复的 https://github.com/manateelazycat/snails/commit/44f77cc35f75e623a6e2a419629a16ec9b9c3fbf 问题，以为是 snails 的问题，今天回想起来本质是 EAF 的问题

4. 基于上面的逻辑推理，eaf-monitor-key-event 在 https://github.com/manateelazycat/emacs-application-framework/blob/901f2e317eebc92424acadedf87feb04648bd4dd/eaf.el#L388 被调用了

5. 我认为是 https://github.com/manateelazycat/emacs-application-framework/pull/96/files 这个补丁引起的，为了显示 C-h m 的帮助，错误的借用了 eaf-monitor-key-event 的作用导致的

clemera:
[...]

PR #112, opened 2019-12-07
Continue normally for non eaf bindings

clemera:
I found that after pre-command-hook is done running the command executes anyway so there is no need to call the command from the hook. This was another reason commands were called twice (see #110).

I also removed resetting last-command-event since self-insert-commands are captured before and there should be no reason for it, I believe.

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -509,7 +509,7 @@ Please ONLY use `eaf-bind-key' to edit EAF keybindings!"
 
 (defun eaf-monitor-key-event ()
   "Monitor key events during EAF process."
-  (ignore-errors
+  (ignore-errors ;so it does not get removed from `pre-command-hook'
     (let* ((event last-command-event)
            (key (if event
                     (make-vector 1 event)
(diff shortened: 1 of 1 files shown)


MatthewZMD:
Check if it works in eaf-browser, inputting arbitrary command in ddg or google

clemera:
Yes, normal input works.

clemera:
Oh for some reason I can not input an exclamation mark did you mean that?

MatthewZMD:
Exclamation mark and some other symbols are yet to be supported, that's okay

MatthewZMD:
> Exclamation mark and some other symbols are yet to be supported

Of course, they should be supported... but not something you introduced in this commit I suppose

clemera:
Oh I noticed that `last-command-event` was reset before because apparently the characters get inserted into the backing buffer and some time ago this buffer was read only which explains the comment that the read only messages should not be displayed. 

Please don't merge this yet. I will investigate a bit.

clemera:
Okay know I understand things a bit better. For commands dispatched to `eaf-call` the real emacs still gets called in addition to `eaf-call`. So for self-insert-command the chars are added to the emacs buffer:
    
        (with-current-buffer "... DuckDuckGo"
      (buffer-string)) ;; => input in search bar

Probably when the command is handled by eaf the command loop should be stopped. This could be done by throwing an error but erros are ignored so the function does not get removed from pre-command-hook. Man that's getting complicated.

clemera:
One could override the command beeing execute one time with a command which just removes the advice again. A bit crazy but I tested it and it would work.

MatthewZMD:
Looks very hacky, haha. Wait until Andy has woke up to decide.

I won't have any time to hack around emacs for around 2 weeks :(

clemera:
Yes it's hacky, probably not a good idea but it works ;)

manateelazycat:
I send new patch https://github.com/manateelazycat/emacs-application-framework/commit/bcf6531d50321003c20190beb73376910d07753f to simple the implementation of eaf-monitor-key-event.

@clemera Thanks for help!

manateelazycat:
Patch https://github.com/manateelazycat/emacs-application-framework/commit/dd647100adf71c6cf47a2498a4e7d3dd9d965097 add support to input exclamation mark 

Thank you guys. @clemera @MatthewZMD

PR #114, opened 2019-12-08
Allow to execute app cmds programmatically

clemera:
This enables to invoke python commands like `scroll_down` and others from `M-x` and similar. A new function for invoking python commands from lisp code was also added which can be used like:

    (eaf-execute-app-cmd 'scroll_down "eaf-buffer-name...")

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -391,7 +391,17 @@ We need calcuate render allocation to make sure no black border around render co
           (random (expt 16 4))
           (random (expt 16 4))))
 
-(defun eaf-dummy-function (sym)
+(defun eaf-execute-app-cmd (cmd &optional buf)
+  "Execute app CMD.
+
+If BUF is given it should be the eaf buffer for the command
+otherwise it is assumed that the current buffer is the eaf
+buffer."
+  (with-current-buffer (or buf (current-buffer))
+    (let ((this-command cmd))
+      (call-interactively cmd))))
+
+
 (defun eaf-dummy-function (sym key)
   "Define an alias from SYM to a dummy function that acts as a placeholder."
   (defalias sym (lambda nil
--- a/eaf.el
+++ b/eaf.el
@@ -392,20 +392,31 @@ We need calcuate render allocation to make sure no black border around render co
           (random (expt 16 4))))
 
 (defun eaf-dummy-function (sym)
+(defun eaf-dummy-function (sym key)
   "Define an alias from SYM to a dummy function that acts as a placeholder."
   (defalias sym (lambda nil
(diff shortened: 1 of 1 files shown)


manateelazycat:
what purpose of this patch?
I dont know how to use this feature?

clemera:
This is to be able to call `scroll_down` and other commands from `M-x` and from lisp code. With the patch you can go to your app buffer and call the commands from `M-x`

clemera:
Before when doing `M-x scroll_up` nothing would happen because `pre-command-hook` does not run for executing `scroll_up` command.

manateelazycat:
I don't think it's a right tactics to implement this feature.

If you want call Python function from Elisp side, you can build your own interactive function to wrap ```(eaf-call "execute_function" buffer-id function-name)```

It's nothing need change eaf-dummy-function, you should not modify eaf-monitor-key-event function.

All function name in eaf buffer local map is function name on Python side, those Python side function name will conflict with elisp function name provide by other elisp package.

clemera:
Okay, but I think it would be nice to be able to call the commands from M-x like regular Emacs commands. What if you don't remember the key? With a completion framework you can now search for the command and execute it.

>  those Python side function name will conflict with elisp function name provide by other elisp package

This is the case anyway, maybe all those dummy functions should get `eaf` prefix? So we have `eaf-scroll_down` and so on.

manateelazycat:
> Okay, but I think it would be nice to be able to call the commands from M-x like regular Emacs commands. What if you don't remember the key? With a completion framework you can now search for the command and execute it.
> 
> > those Python side function name will conflict with elisp function name provide by other elisp package
> 
> This is the case anyway, maybe all those dummy functions should get `eaf` prefix? So we have `eaf-scroll_down` and so on.

Add eaf- prefix is good idea, it will not conflict with Elisp function.
But i think we should call ```(eaf-call "execute_function" buffer-id function-name)``` in eaf-dummy-function, not should call eaf-monitor-key-event, eaf-monitor-key-event is use for pre-command-hook callback

clemera:
> we should call (eaf-call "execute_function" buffer-id function-name) in eaf-dummy-function, not should call eaf-monitor-key-event

I thought we could reuse it for that purpose but it think you are right it's better to not reuse it.

manateelazycat:
Thanks for patch, i have merge this patch.

clemera:
Should I work on having `eaf` prefix for dummy functions or are you are doing it already?

manateelazycat:
> Should I work on having `eaf` prefix for dummy functions or are you are doing it already?

Yes, welcome to new PR

PR #116, opened 2019-12-08
Refactor namings

clemera:


MatthewZMD:
Coincidentally, there is another patch from Stefen Monnier that will be resolving some of the prefix renames, I am about to create a PR for him (he doesn't use Github) but I don't have enough time atm. 

Since you have already done this, can we set the naming convention for prefixes to `eaf--` instead of `eaf-` for internally-used stuff?

eaf.el
```diff
@@ -94,6 +94,9 @@
     map)
   "Keymap used by `eaf-mode'.")
 
+(defvar-local eaf-buffer-id nil
```
MatthewZMD: As per comment, change to `eaf--buffer-id` instead

MatthewZMD:
Also, on the topic of this, can you rename `buffer-url` into `eaf--buffer-url` and `buffer-app-name` into `eaf--buffer-app-name` too?

clemera:
> can we set the naming convention for prefixes to eaf-- instead of eaf- for internally-used stuff?

+1, I followed the present naming conventions, I will do the other renames, too.

manateelazycat:
Thanks for patch.

PR #120, opened 2019-12-08
Custom group patch from Stefan Monnier

MatthewZMD:
PR representing Stefan Monnier whom sent me the patch through email

- Create a Custom group (you have defcustoms which refer to group `eaf`
  but I couldn't find any corresponding definition of the group).
- Remove the now redundant :group args (vars are put into the last
  group defined by default).

More will be coming... when I have time ;)

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -84,10 +84,14 @@
 (require 'subr-x)
 
 ;;; Code:
+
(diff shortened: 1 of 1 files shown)


manateelazycat:
Thanks for patch. ;)

PR #121, opened 2019-12-09
Code cleanup and Refactors by Stefan Monnier

MatthewZMD:
PR representing Stefan Monnier whom sent me the patch through email

- Prefer derived-mode-p to testing major-mode.
- Quote some vars in docstrings so they appear as hyperlinks in *Help*.
- Remove single-backslashes from strings when they don't do anything
  (e.g. `\/\/` is the same string as `//`).
- A few eta-like rewrites to reduce some code duplication.

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -287,7 +287,7 @@ Any new app should add the its name and the corresponding
 keybinding variable to this list.")
 
 (defun eaf-call (method &rest args)
-  (apply 'dbus-call-method
+  (apply #'dbus-call-method
          :session                   ; use the session (not system) bus
          "com.lazycat.eaf"          ; service name
          "/com/lazycat/eaf"         ; path name
@@ -298,12 +298,12 @@ keybinding variable to this list.")
   (frame-parameter frame 'window-id))
 
 (defun eaf-start-process ()
-  "Start EAF process if it haven't started yet."
+  "Start EAF process if it hasn't started yet."
   (interactive)
   (if (process-live-p eaf-process)
       (message "EAF process has started.")
     (setq eaf-process
-          (apply 'start-process
+          (apply #'start-process
                  eaf-name
                  eaf-name
                  eaf-python-command (append (list eaf-python-file) (eaf-get-render-size) (list eaf-http-proxy-host eaf-http-proxy-port))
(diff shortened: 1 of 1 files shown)


eaf.el
```diff
@@ -591,35 +589,33 @@ Use it as (eaf-bind-key var key eaf-app-keybinding)"
                        (h (nth 3 window-allocation))
                        )
                   (when (and
-                         (> mouse-press-x x)
-                         (< mouse-press-x (+ x w))
-                         (> mouse-press-y y)
-                         (< mouse-press-y (+ y h)))
+                         (< x mouse-press-x (+ x w))
+                         (< y mouse-press-y (+ y h)))
```
manateelazycat: Why remove ```(> mouse-press-x x)``` and ```(> mouse-press-y y)``` ?

eaf.el
```diff
@@ -549,7 +547,7 @@ Please ONLY use `eaf-bind-key' to edit EAF keybindings!"
         ;; Send key to Python side if key-command is single character key.
         ((or (equal key-command "self-insert-command")
              (equal key-command "completion-select-if-within-overlay")
-             (equal key-command "nil")
+             (eq key-command 'nil)
```
manateelazycat: symbol-name always return "nil" string, 
Have you test this change?

MatthewZMD: I tested it and don't see any problem. `eq` compares the object `nil` and `key-command` right? If you feel unsafe, I can change it back.

eaf.el
MatthewZMD: `(and (> mouse-press-y y) (< mouse-press-y (+ y h)))` is the same as `(< x mouse-press-x (+ x w))`

eaf.el
manateelazycat: I'm sure symbol-name always return string (eq "nil" 'nil) is nil , it's wrong

eaf.el
manateelazycat: Correctly logic is:

                         (> mouse-press-x x)
                         (< mouse-press-x (+ x w))
                         (> mouse-press-y y)
                         (< mouse-press-y (+ y h))

manateelazycat: Oh, I see.

< support many params

manateelazycat:
@MatthewZMD I push new https://github.com/manateelazycat/emacs-application-framework/commit/6b63a92594883b166d09bca615273b095bb921b6 fix eq problem.

Issue #117, opened 2019-12-08
Implement eaf apps as minor-modes


clemera:
Currently `eaf-mode-map` is used by every app. This means if you use multiple apps all have the keymap from the last started app, which is not good ;)

Another problem with the current approach is that if the user has some minor mode which bindings conflict with those of the eaf-mode-map the eaf bindings have no precedence and the minor mode bindings are used.

I think implementing apps as minor modes would be a good idea. Then users also could change bindings by just changing `eaf-browser-mode-map` for example. What do you think? Should I start working on this?

MatthewZMD:
When I was working on implementing C-h m with dummy variables, I also thought able organizing them with minor modes. However it turns out not looking very good on C-h m -- because all specific keybindings will be somewhere burried under the Minor mode spec ocean, not very helpful for someone wanting to quickly check keybindings by C-h m

clemera:
> However, it turns out not looking very good on C-h m

An alternative I can think of would be to use `C-h v eaf-browser-mode-map` to check the bindings. When executing a command using `counsel-M-x` and I believe `helm` bindings are shown inline, too. A custom `describe-eaf-bindings` would also be possible.

> I can see this issue in a lot of emacs packages, that some minor mode keybinding overrides a major mode keybinding.

I think when inside an app the eaf bindings should have precedence over all other bindings the user has currently defined. I see no reasons users would define other minor-modes for eaf buffers so there would be no conflict in this case. Having a function to change the bindings is nice but it's additional work we could save users from.

clemera:
> We could replace the keymaps depending on the current EAF app buffer the user is currently in instead of replacing them during app start-up, that's a possible solution.

If we should decide doing that we could define a buffer local `eaf-mode-map` in each app buffer and then set the default value to the buffer local value the user is currently in. I'm still in favor for using minor-modes, though ;)

clemera:
One other solution I just thought of to have always precedence for the app keys would be to use a buffer local `overriding-terminal-local-map` but that may introduce other problems, I'm not aware of.

manateelazycat:
I personally do not care about this feature, because I basically do not see the key details. (Perhaps because I was too familiar with this project, hahahaha)

Anyway, any improvement are welcome! ;)

MatthewZMD:
Well, we can also ask the person who raised this feature request at #82 @stardiviner for opinions on this topic

stardiviner:
[...]

PR #124, opened 2019-12-09
New keymap features and fix some key problems

clemera:
This adds a default map which can be used for bindings which you can use in all apps. This also adds a command bound in this map to fix problem with `C-h m` (as discussed in #117)  when using multiple apps in parallel.

If you want to bind other keys which should be available for all apps use:

    (define-key eaf-mode-map* ...)

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -450,7 +450,9 @@ Please ONLY use `eaf-bind-key' to edit EAF keybindings!"
          (eaf-buffer (generate-new-buffer (truncate-string-to-width file-or-command-name eaf-title-length))))
     (with-current-buffer eaf-buffer
       (eaf-mode)
-      (make-local-variable 'emulation-mode-map-alists)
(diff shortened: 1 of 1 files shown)


clemera:
The last commit also fixes the precedence problem, for example currently there is following problem, you can test as follows:

    (defvar test-mode-map
      (let ((map (make-sparse-keymap)))
        (define-key map (kbd "M-b") 'previous-buffer)
        map))

    (define-minor-mode test-mode
      "Haha"
      :global t
      :keymap test-mode-map)

    (test-mode 1)

With the above setup M-b in browser app would navigate back but  *also* switch to previous buffer. To avoid this the last commit adds buffer local emulation-mode-alist which fixes this problem.

manateelazycat:
Thanks for help

MatthewZMD:
Awesome!

PR #128, opened 2019-12-10
Use advice-add instead of the old defadvice by Stefan Monnier

MatthewZMD:
PR representing Stefan Monnier who sent me the patch through email

He noted this too, what do you think
```
;; FIXME: In the code below we should use `save-selected-window' (or even
;; better `with-selected-window') rather than (other-window +1) followed by
;; (other-window -1) since this is not always a no-op.
```


manateelazycat:
need test

MatthewZMD:
I've tested the `device-add` patches without seeing any issues, to convert from `other-window` to `save-selected-window` as per FIXME we need to write code to try it out. I don't have a lot of time right now, can you do it?

manateelazycat:
leave it, we improve it in the future

MatthewZMD:
OK! :+1:

PR #130, opened 2019-12-10
`key-command` is unnecessary, `this-command` handles the job.

MatthewZMD:
Also, Use `member` to compare commands


MatthewZMD:
@manateelazycat 懒猫，有空看一下emacs-china私信吧 ;)

PR #132, opened 2019-12-10
Add `eaf-capture-commands`

clemera:
Commands like `self-insert-command` which send keys to the python side still run on the Emacs side, too. For example check the buffer contents of eaf buffer after typing something in browser app:

    (with-current-buffer "<eaf browser buffer" (buffer-string)) ;; => the content you typed

This patch ensures those commands are ignore on the Emacs side. So characters are not inserted into the eaf buffer in the background. I also added `delete-backward-char` so you don't get beginning of buffer error when the command runs in the eaf buffer.


manateelazycat:
awesome

Issue #62, opened 2019-07-06
Suggestion: save eaf buffers in bookmarks


MatthewZMD:
I really enjoy the speed and capabilities of eaf. However, it would be nicer and more user-friendly if we can save eaf buffers in bookmarks (or something similar to bookmarks)
Examples:
* Browser: Save Google homepage in a bookmark so we don't need to manually type `eaf-open-url RET https://www.google.com` every time when we need to open google.
* PDF Viewer: Save pdf files in a bookmark. When opening a pdf file directly from a bookmark, it goes to the position we saved last time, without manually finding the file. (similar behavior exists in pdf-tools already)
* Same idea goes with video player, image viewer ...

manateelazycat:
Good idea, PR we welcome!

manateelazycat:
@MatthewZMD 按照你对 EAF 的熟悉程度，我觉得这个功能你可以直接发PR哈，我最近超急忙，估计最多只有时间合并你的补丁了，哈哈哈哈

MatthewZMD:
哈哈哈，我有空时研究研究

stardiviner:
This also can be added to Org Mode link support. Will be very useful for Org Mode.

luhuaei:
我觉得利用[interleave](https://github.com/rudolfochrist/interleave)这种形式代替进行交互可能更加方便。具有书签的功能同时也具有做笔记的功能。

对interleave进行尝试，采用eaf-pdf-viewer接口，可以得到以下的效果，勉强可以用...
![GIFrecord_2019-10-26_141048](https://user-images.githubusercontent.com/46041665/67622794-82615600-f850-11e9-938a-5442986f41ae.gif)可以实现上面所讨论的根据笔记对pdf页面进行跳转。

同样也可以用于浏览器、当浏览某一个书签时，还可以提示你在当前页面是否具有笔记等等。

stardiviner:
This is an good idea!! @luhuaei , how do you implemented that? Can you share your patch code? or add PR to `eaf`? Style like interleave is really awesome, and portable.

luhuaei:
跳转是使用`pdf-viewer`自带的跳转功能，插入笔记是从`pdf-viewer`返回信息，再插入到`org`中。你可以参考 https://github.com/luhuaei/interleave/blob/add-eaf/interleave.el  @stardiviner

MatthewZMD:
[...]

PR #133, opened 2019-12-10
Add bookmark feature for eaf browser

clemera:
This adds bookmark as discussed in #62 . For now only browser app is supported. I have shortly tested it and it work but needs more testing, I guess.

To test: `C- x r m` in browser app will add a bookmark. You can then open via `C-x r l` or whatever you use to open bookmarks (`helm-bookmarks`, `counsel-bookmark` or maybe snails ;)

### DIFF
--- a/core/browser_buffer.py
+++ b/core/browser_buffer.py
@@ -91,8 +91,7 @@ class BrowserBuffer(Buffer):
     def scroll_to_bottom(self):
         self.eval_js("window.scrollBy(0, document.body.scrollHeight)")
 
-    def set_bookmark(self):
-        url = self.buffer_widget.web_page.executeJavaScript("window.location.href;")
-        self.buffer_widget.set_emacs_var.emit("eaf--browser-current-url", url)
+    def get_bookmark(self):
+        return self.buffer_widget.web_page.executeJavaScript("window.location.href;")
 
 
--- a/core/browser_buffer.py
+++ b/core/browser_buffer.py
@@ -90,3 +90,9 @@ class BrowserBuffer(Buffer):
 
     def scroll_to_bottom(self):
         self.eval_js("window.scrollBy(0, document.body.scrollHeight)")
+
+    def set_url_for_bookmark(self):
+        url = self.buffer_widget.web_page.executeJavaScript("window.location.href;")
+        self.buffer_widget.set_emacs_var.emit("eaf--browser-current-url", url)
+
+
--- a/core/browser_buffer.py
+++ b/core/browser_buffer.py
@@ -91,7 +91,7 @@ class BrowserBuffer(Buffer):
     def scroll_to_bottom(self):
         self.eval_js("window.scrollBy(0, document.body.scrollHeight)")
 
-    def set_url_for_bookmark(self):
+    def set_bookmark(self):
         url = self.buffer_widget.web_page.executeJavaScript("window.location.href;")
         self.buffer_widget.set_emacs_var.emit("eaf--browser-current-url", url)
 
--- a/core/buffer.py
+++ b/core/buffer.py
@@ -271,5 +271,5 @@ class Buffer(QGraphicsScene):
     def update_settings(self):
         pass
 
-    def set_bookmark(self):
+    def get_bookmark(self):
         pass
--- a/core/buffer.py
+++ b/core/buffer.py
@@ -272,4 +272,4 @@ class Buffer(QGraphicsScene):
         pass
 
     def get_bookmark(self):
-        pass
+        return ""
--- a/core/buffer.py
+++ b/core/buffer.py
@@ -267,3 +267,6 @@ class Buffer(QGraphicsScene):
 
     def update_settings(self):
         pass
+
+    def set_bookmark(self):
+        pass
--- a/eaf.el
+++ b/eaf.el
@@ -329,6 +329,25 @@ For now only eaf browser app is supported."
     (cond ((equal app "browser")
            (eaf-open-url (cdr (assq 'filename bookmark)))))))
 
+(defun eaf-browser-open-bookmark ()
+  "Command to open or create eaf bookmarks with completion."
+  (interactive)
+  (bookmark-maybe-load-default-file)
+  (let* ((bookmarks (cl-remove-if-not
+                     (lambda (entry)
+                       (bookmark-prop-get entry 'eaf-app))
+                     bookmark-alist))
+         (names (mapcar #'car bookmarks))
+         (cand (completing-read "Eaf bookmark: " bookmarks)))
+    (cond ((member cand names)
+           (bookmark-jump cand))
+          (t
+           (unless (derived-mode-p 'eaf-mode)
+             (user-error "Not in an eaf buffer"))
+           ;; create new one
+           (bookmark-set cand)))))
+
+
 (defun eaf-call (method &rest args)
   (apply #'dbus-call-method
          :session                   ; use the session (not system) bus
--- a/eaf.el
+++ b/eaf.el
@@ -304,8 +304,8 @@ Any new app should add the its name and the corresponding
 keybinding variable to this list.")
 
 
-(defvar eaf--browser-current-url nil)
-(defvar-local eaf--full-title nil)
+(defvar eaf--bookmark-link nil)
+(defvar-local eaf--bookmark-title nil)
 
 (defun eaf--bookmark-make-record ()
   "Create a eaf bookmark.
@@ -313,15 +313,15 @@ keybinding variable to this list.")
 The bookmark will try to recreate eaf buffer session.
 For now only eaf browser app is supported."
   (cond ((equal eaf--buffer-app-name "browser")
-         ;; set `eaf--browser-current-url'
+         ;; set `eaf--bookmark-link'
          (eaf-call "execute_function" eaf--buffer-id
                    "set_bookmark")
          (let ((bookmark `((handler . eaf--bookmark-restore)
                            (eaf-app . "browser")
-                           (defaults . ,(list eaf--full-title))
+                           (defaults . ,(list eaf--bookmark-title))
                            ;; not a filename but this shows url in bookmark-list
                            ;; which is nice
-                           (filename . ,eaf--browser-current-url))))
+                           (filename . ,eaf--bookmark-link))))
            bookmark))))
 
 (defun eaf--bookmark-restore (bookmark)
@@ -342,8 +342,8 @@ For now only eaf browser app is supported."
          (cand (completing-read "Eaf bookmark: "
                                 bookmarks
                                 nil nil nil nil
-                                (unless (member eaf--full-title names)
-                                  (format "+%s" eaf--full-title)))))
+                                (unless (member eaf--bookmark-title names)
+                                  (format "+%s" eaf--bookmark-title)))))
     (cond ((member cand names)
            (bookmark-jump cand))
           (t
@@ -744,7 +744,7 @@ Use it as (eaf-bind-key var key eaf-app-keybinding)"
             (when (and
                    (derived-mode-p 'eaf-mode)
                    (equal eaf--buffer-id bid))
-              (setq-local eaf--full-title title)
+              (setq-local eaf--bookmark-title title)
               (rename-buffer (truncate-string-to-width title eaf-title-length))
               (throw 'find-buffer t))))))))
 
--- a/eaf.el
+++ b/eaf.el
@@ -130,9 +130,11 @@ Don't modify this map directly. To bind keys for all apps use
   ;; which may not want this, introduce eaf user option?
   (setq window-combination-resize t)
   (set (make-local-variable 'eaf--buffer-id) (eaf-generate-id))
+  (setq-local bookmark-make-record-function 'eaf--bookmark-make-record)
   ;; Fix #110 , make `eaf-monitor-key-event' buffer locally to pre-command-hook of the eaf-mode buffer.
   ;; To fix interactive command run twice because `eaf-monitor-key-event' runs inside minibuffer and can not handle minibuffer quit signal.
   (add-hook 'pre-command-hook #'eaf-monitor-key-event nil t))
+
 
 (defvar eaf-python-file (expand-file-name "eaf.py" (file-name-directory load-file-name)))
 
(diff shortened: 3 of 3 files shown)


clemera:
I have now added a command to open/create bookmarks with completion, too.

core/browser_buffer.py
```diff
@@ -90,3 +90,9 @@ def scroll_to_begin(self):
 
     def scroll_to_bottom(self):
         self.eval_js("window.scrollBy(0, document.body.scrollHeight)")
+
+    def set_url_for_bookmark(self):
```
manateelazycat: We should define interface function
```
def set_bookmark(self):
    pass
```

in emacs-application-framework/core/buffer.py

And implement interface ```set_bookmark``` in core/browser_buffer.py

eaf.el
```diff
@@ -687,6 +744,7 @@ Use it as (eaf-bind-key var key eaf-app-keybinding)"
             (when (and
                    (derived-mode-p 'eaf-mode)
                    (equal eaf--buffer-id bid))
+              (setq-local eaf--browser-full-title title)
```
manateelazycat: Not just browser title, we should use eaf--full-title

manateelazycat:
@clemera Thanks for patch.

But i think we should commit this patch with interface way, not just browser.

Within this patch we can only implement browser application interface, but we need to create a bookmark interfaces, then implementation of the interface to do,
otherwise late to bookmark other applications to support a lot of trouble, bad maintenance.

Thanks for patch, I think it's good start to support bookmark.

eaf.el
```diff
@@ -97,6 +97,7 @@
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "C-h m") 'eaf-describe-bindings)
     (define-key map [remap describe-bindings] 'eaf-describe-bindings)
+    (define-key map (kbd "C-c b") 'eaf-browser-open-bookmark)
```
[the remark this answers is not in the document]
clemera: Good catch, thanks!

core/browser_buffer.py
clemera: I'm not as fluent on the python side can you do it?

eaf.el
clemera: Okay

clemera:
What kind of interface would you like to create? I dispatch in cond of `eaf--bookmark-make-record` and `eaf--bookmark-restore` based on app name. Should another alist be added to map app names to bookmark handlers?

core/browser_buffer.py
manateelazycat: Example, search ```scroll``` interface in core/buffer.py, you can define an empty interface in core/buffer.py and don't need implement it, then you can call ```scroll``` interface in Elisp side with eaf-call "execute_function".

And check app/pdf-viewer file, you can found ```scroll``` implement in pdf-viewer/buffer.py

In bookmark, you can define 
```
def set_bookmark(self):
    pass
```

in core/buffer.py 

Then implmenet ```set_bookarm``` in browser_buffer.py as your current patch content.

eaf.el
```diff
@@ -300,6 +303,59 @@ Try not to modify this alist directly. Use `eaf-setq' to modify instead."
...
+
+The bookmark will try to recreate eaf buffer session.
+For now only eaf browser app is supported."
+  (cond ((equal eaf--buffer-app-name "browser")
+         ;; set `eaf--browser-current-url'
+         (eaf-call "execute_function" eaf--buffer-id
+                   "set_bookmark")
+         (let ((bookmark `((handler . eaf--bookmark-restore)
+                           (eaf-app . "browser")
+                           (defaults . ,(list eaf--full-title))
```
manateelazycat: eaf--full-title should rename to eaf--bookmark-title

eaf.el
```diff
@@ -300,6 +303,59 @@ Try not to modify this alist directly. Use `eaf-setq' to modify instead."
...
+                           ;; not a filename but this shows url in bookmark-list
+                           ;; which is nice
+                           (filename . ,eaf--browser-current-url))))
```
manateelazycat: Change eaf--browser-current-url to eaf--bookmark-link

eaf.el
```diff
@@ -300,6 +303,59 @@ Try not to modify this alist directly. Use `eaf-setq' to modify instead."
 Any new app should add the its name and the corresponding
 keybinding variable to this list.")
 
+
+(defvar eaf--browser-current-url nil)
```
manateelazycat: Change to (defvar-local eaf--bookmark-link nil)

eaf.el
```diff
@@ -300,6 +303,59 @@ Try not to modify this alist directly. Use `eaf-setq' to modify instead."
...
+(defvar-local eaf--full-title nil)
```
manateelazycat: Change to eaf--bookmark-title

manateelazycat:
@clemera We can rename var to eaf--bookmark-title and eaf--bookmark-link

In browser app, eaf--bookmark-link is url string, such as, https://www.google.com
In Pdf Viewer app, eaf--bookmark-link is file://home/andy/test.pdf

Maybe we should add eaf--bookmark-additional-info, such as "39", then we can use eaf--bookmark-link open Pdf file file://home/andy/test.pdf , then jump to page 39 by eaf--bookmark-additional-info variable.

manateelazycat:
Go out for an hour, reply later. ;)

eaf.el
clemera: Done

eaf.el
clemera: Not sure about this one, the variable gets before making a bookmark and there is no need to keep it buffer local because it just records the last bookmark link used.

eaf.el
clemera: Done

eaf.el
clemera: Done

clemera:
> Maybe we should add eaf--bookmark-additional-info, such as "39", then we can use eaf--bookmark-link open Pdf file file://home/andy/test.pdf , then jump to page 39 by eaf--bookmark-additional-info variable.

We can include any value in the alist we create in `eaf--bookmark-make-record` for the corresponding app.

eaf.el
```diff
@@ -300,6 +303,60 @@ Try not to modify this alist directly. Use `eaf-setq' to modify instead."
 Any new app should add the its name and the corresponding
 keybinding variable to this list.")
 
+
+(defvar eaf--bookmark-link nil)
```
manateelazycat: This should be defvar-local ?

clemera:
I think we should search for a better way to associate information with individual apps. For example we have `eaf-browser-keybinding`and `eaf-browser-extension-list` and now we need to add something like `eaf-bookmark-handler-alist`. Better would be to have something like a stuct for each app with fields describing additional info about it. What do you think?

eaf.el
clemera: Here is my comment from above:

Not sure about this one, the variable gets set before making a bookmark and there is no need to keep it buffer local because it just records the last bookmark link used. We always need the get a fresh value from the python side before using it.

clemera: Thinking about it, maybe we should give it a more general name because we might have other purposes to get values from the python side. This is only for transmitting some info about current state on python side so maybe we should call it something like `eaf--python-state-val`?

manateelazycat:
> I think we should search for a better way to associate information with individual apps. For example we have `eaf-browser-keybinding`and `eaf-browser-extension-list` and now we need to add something like `eaf-bookmark-handler-alist`. Better would be to have something like a stuct for each app with fields describing additional info about it. What do you think?

Yes, alist is neat solution, looking forward it. ;)

clemera:
> Yes, alist is neat solution, looking forward it

You mean one big alist for all apps? So we would have something like:

    (("browser"  . (("extension-list" . ...)
                           ...))
    (("pdf-viewer" . ...)))

eaf.el
manateelazycat: If you just want a temp var, you can check ```new_buffer``` interface define in eaf.py

You can return temp variable by DBus method, just need define out_signature="s" in Dbus header, mean return a string value from Dbus method.

clemera:
I meant a general solution above not only for bookmark handler. We need to associate more info in the future with different apps so I think we should somehow pack them all together so we don't have to create so many variables. 

Then we could have a general function to get some mapping for current app:

    (eaf-get-settings-for-current-app "extension-list" )
    (eaf-get-settings-for-current-app "bookmark-handler" )

and so on.

manateelazycat:
> I meant a general solution above not only for bookmark handler. We need to associate more info in the future with different apps so I think we should somehow pack them all together so we don't have to create so many variables.
> 
> Then we could have a general function to get some mapping for current app:
> 
> ```
> (eaf-get-settings-for-current-app "extension-list" )
> (eaf-get-settings-for-current-app "bookmark-handler" )
> ```
> 
> and so on.

Maybe we can refactory those code in next patch? ;)

clemera:
> Maybe we can refactory those code in next patch

Yes, just wanted to know if you have any opinions on that.

eaf.el
clemera: I did not knew that but that sound like something I want to do ;) But I have no idea how to do that from what you said. Could you give a more detailed example what I would need to do. Here is how the value is currently retrieved:

    def set_bookmark(self):
        url = self.buffer_widget.web_page.executeJavaScript("window.location.href;")
        self.buffer_widget.set_emacs_var.emit("eaf--browser-current-url", url)

How to do that in a way you described?

manateelazycat:
> > Maybe we can refactory those code in next patch
> 
> Yes, just wanted to know if you have any opinions on that.

Wait my next patch, add new interface, you can call function like execute_function interface, but can return temp variable from Python side, then you don't need define temp variable like eaf--bookmark-link

core/browser_buffer.py
```diff
@@ -90,3 +90,9 @@ def scroll_to_begin(self):
 
     def scroll_to_bottom(self):
         self.eval_js("window.scrollBy(0, document.body.scrollHeight)")
+
+    def set_bookmark(self):
+        url = self.buffer_widget.web_page.executeJavaScript("window.location.href;")
+        self.buffer_widget.set_emacs_var.emit("eaf--browser-current-url", url)
+
```
manateelazycat: Check patch https://github.com/manateelazycat/emacs-application-framework/commit/465c3f3ec4a631932477e5e3e9f1bb77c62d9ba8

Change set_bookmark to get_bookmark:

def get_bookmark(self):
    return self.buffer_widget.web_page.executeJavaScript("window.location.href;")

core/buffer.py
```diff
@@ -267,3 +267,6 @@ def fake_key_event(self, event_string):
 
     def update_settings(self):
         pass
+
+    def set_bookmark(self):
```
manateelazycat: Change to get_bookmark

eaf.el
```diff
@@ -300,6 +303,60 @@ Try not to modify this alist directly. Use `eaf-setq' to modify instead."
...
+
+(defun eaf--bookmark-make-record ()
+  "Create a eaf bookmark.
+
+The bookmark will try to recreate eaf buffer session.
+For now only eaf browser app is supported."
+  (cond ((equal eaf--buffer-app-name "browser")
+         ;; set `eaf--bookmark-link'
+         (eaf-call "execute_function" eaf--buffer-id
+                   "set_bookmark")
```
manateelazycat: Remove those code

eaf.el
```diff
@@ -300,6 +303,60 @@ Try not to modify this alist directly. Use `eaf-setq' to modify instead."
...
+         (let ((bookmark `((handler . eaf--bookmark-restore)
+                           (eaf-app . "browser")
+                           (defaults . ,(list eaf--bookmark-title))
+                           ;; not a filename but this shows url in bookmark-list
+                           ;; which is nice
+                           (filename . ,eaf--bookmark-link))))
```
manateelazycat: (filename . ,(eaf-call "call_function" eaf--buffer-id "get_bookmar"))

manateelazycat:
@clemera In patch https://github.com/manateelazycat/emacs-application-framework/commit/465c3f3ec4a631932477e5e3e9f1bb77c62d9ba8

I add new DBus method call_function at Python side:
Now you can use (setq temp-var (eaf-call "call_function" buffer-id "get_bookmar")) to return temp var to elisp side, don't need define temp variable everywhere. ;)

core/browser_buffer.py
clemera: Nice, thanks, this dbus stuff is still black magic to me ;)

clemera:
Nice, thanks, this dbus stuff is still black magic to me ;)

core/browser_buffer.py
manateelazycat: Default, execute_function just execute function, call_function return Python function result to Elisp side through DBus IPC, then you don't need set_emacs_var.emit set temp elisp variable everywhere. ;)

eaf.el
```diff
@@ -300,6 +303,57 @@ Try not to modify this alist directly. Use `eaf-setq' to modify instead."
...
+For now only eaf browser app is supported."
+  (cond ((equal eaf--buffer-app-name "browser")
+         (let ((bookmark `((handler . eaf--bookmark-restore)
+                           (eaf-app . "browser")
+                           (defaults . ,(list eaf--bookmark-title))
+                           ;; not a filename but this shows url in bookmark-list
+                           ;; which is nice
+                           (filename . ,(eaf-call "call_function"
+                                                  eaf--buffer-id "get_bookmark")))))
+           bookmark))))
```
manateelazycat: Remove let sexp ?

core/buffer.py
```diff
@@ -270,3 +270,6 @@ def fake_key_event(self, event_string):
 
     def update_settings(self):
         pass
+
+    def get_bookmark(self):
+        pass
```
manateelazycat: This need change to

```Python
def get_bookmark(self):
    return ""
```

Because call_function DBus always need get_bookmark return string.

manateelazycat:
@clemera Awesome job, thank you!!! ;)

clemera:
> Now you can use (setq temp-var (eaf-call "call_function" buffer-id "get_bookmar")) 

Would it be possible to have something equivalent on python side?

    temp_var = emacs_call_fun(...)

I could use that in my color picker app.

clemera:
Oh wait, I did not know that you would merge it, yet. Haven't tested with recent changes.

manateelazycat:
> > Now you can use (setq temp-var (eaf-call "call_function" buffer-id "get_bookmar"))
> 
> Would it be possible to have something equivalent on python side?
> 
> ```
> temp_var = emacs_call_fun(...)
> ```
> 
> I could use that in my color picker app.

Check https://github.com/jwiegley/emacs-release/blob/adfd5933358fdf6715423dee8169eb3cd7d982db/lisp/net/dbus.el#L762 

Same as call_function interface, we can call Elisp function and return reulst to Python side. ;)

manateelazycat:
> Oh wait, I did not know that you would merge it, yet. Haven't tested with recent changes.

PR looks good, welcome to send new PR if you found new bug. ;)

clemera:
Okay thanks!

MatthewZMD:
Waking up to see this being supported now, this is amazing. Thank you!

PR #141, opened 2019-12-11
Re-order dolists, otherwise might overwrite keys in eaf-capture-keys

MatthewZMD:
For example, `eaf-pdf-viewer-keybinding` has `("SPC" . "scroll_up_page")`, it will get overriden after applying
```
(dolist (single-key eaf-capture-keys)
            (define-key map (kbd single-key) 'eaf-send-key))
```
because `SPC` is in `eaf-capture-keys`.

Therefore we rearrange the dolist to the beginning to avoid this issue.


manateelazycat:
Thanks for patch

PR #142, opened 2019-12-11
Some more refactorings

clemera:
Move emulation map init into mode macro and init buffer vars in `eaf-create-buffer`


manateelazycat:
Thanks for patch

PR #143, opened 2019-12-11
Refactoring eaf-open and eaf-open-internal

clemera:
Separate command init code and display code. Add
`eaf-app-display-function-alist` to configure display of apps.


clemera:
Please don't merge it yet I noticed I need to do some more.

PR #144, opened 2019-12-11
Refactor hooks

clemera:
- Don't run window hooks if eaf-proccess not live
- Move hooks from top level


clemera:
Wait until #143 gets merged because save hooks needs to move there.

PR #145, opened 2019-12-12
Refactor emacs hook usage

clemera:
Before adding hooks was done when loading the package now it only happens when starting an app. I also made hooks buffer local where possible.

One thing I noticed when testing this is that org-preview does not sync but that problem existed before. It saves on export but the updated file is not displayed for some reason.

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -136,7 +136,8 @@ Don't modify this map directly. To bind keys for all apps use
   (setq-local emulation-mode-map-alists
               (default-value 'emulation-mode-map-alists))
   (push (list (cons t eaf-mode-map))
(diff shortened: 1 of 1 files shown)


manateelazycat:
Thanks

clemera:
@manateelazycat I noticed org-preview does not sync the view on save:

    (eaf-call "update_buffer_with_url" "app.orgpreviewer.buffer" (buffer-file-name) "")

Seems to fail, but I'm not sure why

manateelazycat:
i dont use org mode

clemera:
Okay, I will try to investigate

Issue #140, opened 2019-12-11
Rename all EAF bindings to have eaf- prefix or don't use it at all


MatthewZMD:
Right now, this line
```
do (let ((dummy (intern (format "eaf-%s" fun))))
```
will put an `eaf-` prefix for all dummy functions. 

There is no functionality issue with this approach, however, the `*Help*` buffer can give misinformation.

For example,
```
eaf-switch_to_read_mode is an interactive Lisp closure.

(eaf-switch_to_read_mode)

This Lisp function is a placeholder, the actual function will be handled on the Python side.

Use ‘eaf-execute-app-cmd’ if you want to execute this command programmatically.
Please ONLY use ‘eaf-bind-key’ to edit EAF keybindings!
```

The user, upon reading the last line, might try to use `eaf-bind-key` function with the variable name  on the first line, which is `eaf-switch_to_read_mode`. This will break because they're supposed to  call `eaf-bind-key` with the variable name `switch_to_read_mode` instead. 

Therefore, I propose we either rename *all* EAF bindings to have a `eaf-` prefix, or we don't format with the prefix at all using `(format "eaf-%s" fun)`

manateelazycat:
Check https://github.com/manateelazycat/emacs-application-framework/commit/2a833fab1991085c33f1ac00d07be2066534276f

PR #146, opened 2019-12-12
Add refresh command for browser app and prefix app commands like before

clemera:


clemera:
I included the app name in the prefix, too. For example pdf and browser have different scroll keys so the dummy command needs to distinguish the key it's bound too.

MatthewZMD:
Please check #140

clemera:
Thanks, I missed that and wondered how the prefix vanished ;) Not sure how to deal with this currently.

clemera:
Another benefit of having prefixes is search: When there are many apps you can input `eaf browser` (assuming you use some completion framework) and see all the commands available for the browser app.

MatthewZMD:
Yes I'm sure. We can also make eaf-bind-key smarter, make it recognize variable names with and without prefix. A simple regex would do that.

clemera:
I have added an explanation mentioning the unprefixed name to the dosctring

MatthewZMD:
Great

PR #148, opened 2019-12-12
Add hook for apps

clemera:
Sometimes the user might want to run app specific setup (for example to activate
minor modes which change the appearance of the mode line and stuff like that). This allows this, all
one need to do is to define a hook variable named `eaf-<app-name>-hook`.


clemera:
General question: How would you like to handle little PRs like this. Should I merge them straight away or do you want to always review them?

MatthewZMD:
For small changes, I think you can merge them yourself

PR #149, opened 2019-12-12
Allow to provide regular emacs commands in app keybinding lists

clemera:
If the key is mapped to a symbol it is an Emacs command otherwise
an app command. For example for pdf-viewer I wanted to bind q to bury-buffer command.

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -505,10 +505,12 @@ to edit EAF keybindings!" fun)))
             (define-key map (kbd single-key) 'eaf-send-key))
           (set-keymap-parent map eaf-mode-map*)
           (cl-loop for (key . fun) in keybinding
-                   do (let ((dummy (intern
-                                    (format "eaf-%s-%s" app-name fun))))
-                        (eaf-dummy-function dummy fun key)
-                        (define-key map (kbd key) dummy))
+                   do (if (symbolp fun)
+                          (define-key map (kbd key) fun)
+                        (let ((dummy (intern
+                                      (format "eaf-%s-%s" app-name fun))))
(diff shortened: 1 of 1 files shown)


stardiviner:
I did this too. Updated binding is great! Thanks.

stardiviner:
@clemera What's the eaf-app-command corresponding to close PDF viewer?
```elisp
(eaf-bind-key 'quit-window "q" '???)
```

clemera:
You can use:

    (eaf-bind-key quit-window "q" eaf-pdf-viewer-keybinding)

I will soon add an `eaf-configure-app` command where you only need to provide an app name to configure a setting (keys, extensions, display function and so on).

clemera:
Oh I think that works by accident and this should not be used because it has side effects. For now you should better use:

        (setf (alist-get "q" eaf-pdf-viewer-keybinding)
          'quit-window)

stardiviner:
This still have problem, when I have not opened pdf viewer, Emacs report void variable `eaf-pdf-viewer-keybinding`.

clemera:
The variable is not defined when the eaf package has not been loaded. You can use:

    (with-eval-after-load 'eaf
      (setf (alist-get "q" eaf-pdf-viewer-keybinding)
            'quit-window))

Or if you use `use-package` you can put it into the `:config` section.

stardiviner:
@clemera Thanks! :)

clemera:
@stardiviner Oh there is another problem you have to provide equal for the test function:

    (with-eval-after-load 'eaf
      (setf (alist-get "q" eaf-pdf-viewer-keybinding nil nil 'equal)
            'quit-window))

PR #151, opened 2019-12-13
Add eaf-configure function to configure all relevant app variables

clemera:
Here is an example how one can use it:

    (eaf-configure "browser"
                   :key-alist '(("C-q" . quit-window))
                   :extensions '("xhtml")
                   :display-function* #'switch-to-buffer-other-window)

I hope that matches how you want to approach things in the future. I think one step further would be to use this function to create the settings that exists now and remove the corresponding defcustoms. The idea is to force users to use this function to configure things. This would allow us to easily change implementations/add things in the future without breaking user config, providing a consistent API to configure things. And for new apps we could automatically create the needed variables by deriving their name from the app name. 

What do you think?

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -674,6 +674,70 @@ This is used to bind key to EAF Python applications.
 Use it as (eaf-bind-key var key eaf-app-keybinding)"
   `(map-put ,eaf-app-keybinding ,key (symbol-name ',var)))
 
+(cl-defun eaf-configure
+    (app &key
+         key-alist extensions hook-functions
+         key-alist* extensions* hook-functions*
+         display-function* bookmark-function*
+         )
+  "Configure eaf APP.
+
+Keywords without a trailing '*' update existing settings with the
+provided values.
+
+When using keywords with a trailing '*' any existing settings are
+replaced by the provided value.
+
+KEY-ALIST/KEY-ALIST* is an alist of keys mapped to bindings for
+APP, see `eaf-app-binding-alist'.
+
+EXTENSIONS/EXTENSIONS* is a list of file extensions which should
+be handled by APP, see also `eaf-app-extensions-alist'.
+
+HOOK-FUNCTIONS/HOOK-FUNCTIONS* is a list of functions to be
+called to initilize the APP
+
+DISPLAY-FUNCTION* is the function used to display the APP buffer,
+see `eaf-app-display-function-alist'.
+
+BOOKMARK-FUNCTION* is a handler function to create a bookmark for
+APP, see `eaf-app-bookmark-handlers-alist'."
+  (when (or key-alist key-alist*)
+    (let ((settings (cdr (assoc app eaf-app-binding-alist))))
+      (cond (key-alist*
+             (setf (symbol-value settings) key-alist*))
+            (settings
+             (cl-loop for (key . fun) in key-alist
+                      do (setf (alist-get key (symbol-value settings)
+                                          nil nil #'equal)
+                               fun))))))
+  (when (or extensions extensions*)
+    (let ((settings (cdr (assoc app eaf-app-extensions-alist))))
+      (cond (extensions*
+             (setf (symbol-value settings) extensions*))
+            (settings
+             (cl-loop for ext in extensions
+                      do (pushnew ext (symbol-value settings)
+                                  :test #'equal))))))
+
+  (when (or hook-functions hook-functions*)
+    (let ((hook (intern (format "eaf-%s-hook" app))))
+      (cond (hook-functions*
+             (setf (symbol-value hook) hook-functions*))
+            (hook-functions
+             (dolist (fun hook-functions)
+               (add-hook hook fun))))))
+
+  (when display-function*
+    (setf (alist-get app eaf-app-display-function-alist
+                     nil nil #'equal)
+          display-function*))
+
+  (when bookmark-function*
+    (setf (alist-get app eaf-app-bookmark-handlers-alist
+                     nil nil #'equal)
+          bookmark-function*)))
+
 (defun eaf-focus-buffer (msg)
   (let* ((coordinate-list (split-string msg ","))
          (mouse-press-x (string-to-number (nth 0 coordinate-list)))


manateelazycat:
> Here is an example how one can use it:
> 
> ```
> (eaf-configure "browser"
>                :key-alist '(("C-q" . quit-window))
>                :extensions '("xhtml")
>                :display-function* #'switch-to-buffer-other-window)
> ```
> 
> I hope that matches how you want to approach things in the future. I think one step further would be to use this function to create the settings that exists now and remove the corresponding defcustoms. The idea is to force users to use this function to configure things. This would allow us to easily change implementations/add things in the future without breaking user config, providing a consistent API to configure things. And for new apps we could automatically create the needed variables by deriving their name from the app name.
> 
> What do you think?

First, thanks for this patch.

IMO, I think best software is out of box and user don't need configure anything.
User only need use it directly after install EAF.

eaf-configure is a developer's perspective that provide flexible interface to user for customize, just like package.el does.

I agree we should provide standard and stable APIs for customize, but eaf-configure make thing more complicated.

From user perspective, the only thing I need is keybinding of EAF applicaiton, I don't care any other option, such as ```extensions```, ```display-function``` ,
these are all looking to good, but for users who value a very small thing.

manateelazycat:
Maybe I am old-school emacser, I am very afraid if I need read many document to customize something.

I prefer when you need to customize something, you can just write one line configure to customize it,
rather than to study various variables of this function,
eaf-configure looked really like helm's customize interface,
the first impression is too complicated.

Above is my point, welcome to discuss.

clemera:
>  I think best software is out of box and user don't need configure anything

I totally agree with that. The reason I wrote `eaf-configure` is mostly for convenience to customize things if you need to, which shouldn't be the standard case because hopefully we have configured things to match users expectations. 

The problem I see with the current approach is that it's hard for the user to know which variable to customize. He has to know that he has to search for `eaf-browser-keybinding` to change keybindings, see [here](https://github.com/manateelazycat/emacs-application-framework/pull/149#issuecomment-565261072) for example. For bookmarks would have to find `eaf-app-bookmark-handlers-alist` and so on.

The idea with `eaf-configure` was to have a single entry point so the user can read its docstring to see how to customize any setting he cares about. 

> I am very afraid if I need read many document to customize something

That is a good point, it is good if things work just the usual Emacs way. We can also keep things as they are and only provide `eaf-configure` for additional convenience.

clemera:
Another idea I have would be to have a command like `eaf-customize-app-setting` which would query for the app name and then for the setting you want to customize and after that open a customize buffer for that setting.

manateelazycat:
> > I think best software is out of box and user don't need configure anything
> 
> I totally agree with that. The reason I wrote `eaf-configure` is mostly for convenience to customize things if you need to, which shouldn't be the standard case because hopefully we have configured things to match users expectations.
> 
> The problem I see with the current approach is that it's hard for the user to know which variable to customize. He has to know that he has to search for `eaf-browser-keybinding` to change keybindings, see [here](https://github.com/manateelazycat/emacs-application-framework/pull/149#issuecomment-565261072) for example. For bookmarks would have to find `eaf-app-bookmark-handlers-alist` and so on.
> 
> The idea with `eaf-configure` was to have a single entry point so the user can read its docstring to see how to customize any setting he cares about.
> 
> > I am very afraid if I need read many document to customize something
> 
> That is a good point, it is good if things work just the usual Emacs way. We can also keep things as they are and only provide `eaf-configure` for additional convenience.

Example, currently ```eaf-app-display-function-alist`` just support markdown-previewer and org-previewer, other EAF application don't need care display function.
If user found option ```display-function``` option in eaf-configure, it's hard to understand "What is display function use for?" and "Why other EAF application don't need display function?"

Another situation, ```eaf-app-bookmark-handlers-alist`` should be internal variable for simple code, user should not care about it.

My point is:

1. For user, user only care about keybinding of EAF application.
2. For hacker, they know how to cusotmize EAF through read source code.
3. If we agree "out of box" principle, user don't need read so many docstring even document is very detailed, less is more.

You and me is developer, eaf-configure maybe better for developer, but it's too complicated for user,
I can not imagine what the user needs to customize addition to custom shortcuts?

clemera:
Okay, I understand that your intention is to support the most common use case best, maybe it's enough to improve `eaf-bind-key` (to require the app name not the keybinding list variable name) and to add an additional `eaf-bind-keys` to bind multiple keys at once.

MatthewZMD:
I agree with Andy that this function is too complicated from the user's perspective. For example, I don't really understand why would a user ever want to mess around with `eaf-app-display-function-alist` or `eaf-app-bookmark-handlers-alist` or `eaf-app-display-function-alist`.

>  use this function to create the settings that exists now and remove the corresponding defcustoms.

While realizing this is too complicated for a user, it might actually a good fit for internal use. We can tidy a lot of code.

MatthewZMD:
> improve eaf-bind-key (to require the app name not the keybinding list variable name) and to add an additional eaf-bind-keys to bind multiple keys at once.

This is a great idea, welcome to PR

clemera:
> why would a user ever want to mess around with eaf-app-display-function-alist

To display the app buffer in other window user might want to use `switch-buffer-other-window`, but if you both feel it shouldn't be included that's fine. You can close this or if you like include it only for internal use.

clemera:
> This is a great idea, welcome to PR

I think maybe `eaf-bind-key` should stay as it is to not break existing config. And then add an additional `eaf-bind-keys` which has these features?

MatthewZMD:
Now thinking of it again, the user really has no idea what `app-name` really is unless they open up the source code. While they can figure out what `eaf-.*-keybinding` by doing `C-h v` (assuming they have a completion framework)

MatthewZMD:
> if you like include it only for internal use.

There is no need to close it right now, let's wait until Andy wakes up to see if he wants it to be for internal use only. You can continue to commit and let us have an idea to what might look like when its fully geared toward internal use

clemera:
> Now thinking of it again, the user really has no idea what app-name really is unless they open up the source code. While they can figure out what eaf-.*-keybinding

Yes, but in this case they have to know to search for "...keybinding" which is not a standard Emacs term. Maybe we should use standard keymaps named `eaf-browser-map` and such. This way the user would find all relevant maps by searching for eaf.*map which would be nice and they could also use whatever they use to bind keys and we don't have to reinvent the wheel ;)

MatthewZMD:
> Maybe we should use standard keymaps named eaf-browser-map and such. This way the user would find all relevant maps by searching for eaf.*map which would be nice and they could also use whatever they use to bind keys and we don't have to reinvent the wheel ;)

Sounds fair, but please resist from making big changes until I finish refactoring and get a PR from Stefan Monnier in, to reduce the amount of possible code conflicts ;)

clemera:
Thanks for the warning, I will wait :) Nice that Stefan has interest in this project he is an Emacs legend.

manateelazycat:
> > why would a user ever want to mess around with eaf-app-display-function-alist
> 
> One use case might be to display the app buffer in other window then the user can use `switch-buffer-other-window`, if you both feel it shouldn't be included that's fine. You can close this or if you like include it only for internal use.

It's hacker perspective, they will figure out how to customize this after read source code.

manateelazycat:
> Now thinking of it again, the user really has no idea what `app-name` really is unless they open up the source code. While they can figure out what `eaf-.*-keybinding` by doing `C-h v` (assuming they have a completion framework)

We have write docs about keybinding in KEYBINDING.md, user don't need read source code

manateelazycat:
> > Now thinking of it again, the user really has no idea what app-name really is unless they open up the source code. While they can figure out what eaf-.*-keybinding
> 
> Yes, but in this case they have to know to search for "...keybinding" which is not a standard Emacs term. Maybe we should use standard keymaps named `eaf-browser-map` and such. This way the user would find all relevant maps by searching for eaf.*map which would be nice and they could also use whatever they use to bind keys and we don't have to reinvent the wheel ;)

We can write those in document, document is enough. ;)

manateelazycat:
> > if you like include it only for internal use.
> 
> There is no need to close it right now, let's wait until Andy wakes up to see if he wants it to be for internal use only. You can continue to commit and let us have an idea to what might look like when its fully geared toward internal use

Hi, thanks for recently contribution, really.

But I dislike complicated code, I think simple always good, not only for users, developers is the same.

If project is out of the box, user will grow up.
If code is simple enough, more hackers will join us. ;)

I need close this PR now, sorry.

MatthewZMD:
> Thanks for the warning, I will wait :) Nice that Stefan has interest in this project he is an Emacs legend.

Done :)

clemera:
> We can write those in document, document is enough. ;)

Users don't always read documents too ;) Currently they need to figure out three things:

1. How to bind a key
2. How to use `eaf-bind-key`
2. How is the key binding variable name for the app

When a new app is added we have to make sure that we don't forget to document the variable name, too. When we would use normal keymaps most user would now how to use `define-key` or could continue to use any packages which simplfy key bindings like `bind-key`. They still would have to figure out the map name but most Emacs users would assume to find keybindings in something like "eaf-..-map" variables. What do you think, if you give an OK I will implement it ;)

MatthewZMD:
> What do you think, if you give an OK I will try to implement it ;)

I just discussed with Andy, yes you can start implementing this feature :)

PR #152, opened 2019-12-13
Change eaf-capture-keys and eaf-capture-commands to defvar, remove delete-backward-char

MatthewZMD:
Is there a specific reason to include `delete-backward-char` in the `eaf-capture-commands`? Won't it be handled by `eaf-capture-keys` already?

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -93,13 +93,13 @@
   "EAF mode hook."
   :type 'hook)
 
-(defcustom eaf-capture-keys
+(defvar eaf-capture-keys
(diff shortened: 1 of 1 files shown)


clemera:
When I included it it was because my system triggered "<backspace>" but I tested and I don't have this problem anymore. I don't know which factors contribute to if "<backspace>" or "DEL" is triggered. I believe there is a similar rule like for "<tab>" and "TAB" where it depends if you have bound "<tab>" in a reachable keymap so I don't think we should include "<backspace>" in capture keys. Probably it's okay to remove this as long nobody complains :)

MatthewZMD:
Okay, I guess I will merge this now.

MatthewZMD:
Whoops, I made a mistake here..

PR #153, opened 2019-12-14
eaf-proxy-function instead of eaf-dummy-function by Stefan Monnier

MatthewZMD:
Rename from eaf-dummy-function since it's not dummy any more.

    Reduce the number of parameters needed in eaf-proxy-function and
    eaf-gen-keybinding-map

    Clearer eaf-bind-key documentation


clemera:
Maybe the proxy command shouldn't have the private prefix  `eaf--proxy` because it's a command which can be called by the user via `M-x`?

MatthewZMD:
I also struggled a bit trying to decide whether I should leave out the private prefix.

PR #161, opened 2019-12-16
Rename search_text to search_term, smarter number of generated keys

MatthewZMD:
1. Rename `search_text` to `search_term` to reduce ambiguity with the function of the same name
2. Smarter number of generated keys, not fixed to 2


manateelazycat:
Thanks

PR #173, opened 2019-12-20
Update eaf bind keys and docstrings

clemera:
I tried to make it clearer for people who look at the mode docstring how to customize bindings and point them to `eaf-bind-key` from there. 

I updated `eaf-bind-key` so one can provide regular commands like:

    (eaf-bind-key quit-window  "q" eaf-pdf-viewer-keybinding)

The implementation assumes app commands always use an underscore, I'm not sure this is the best way to go but that might suffice for now. 

What do you think about additionally indicating the app name in the mode line? Maybe something like:

    <buffer-name>[<app-name>]

Or maybe we could append the app name to the mode-name:

    EAF: <app-name>

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -130,11 +130,16 @@ Don't modify this map directly. To bind keys for all apps use
   "The buffer app name.")
 
 (define-derived-mode eaf-mode fundamental-mode "EAF"
-  "Major mode for Emacs Application Framework.
+  "Major mode for Emacs Application Framework buffers.
 
-This mode is used by all apps. Each app can setup app specific
-hooks by declaring `eaf-<app-name>-hook'. This hook runs after
-the app buffer has initialized."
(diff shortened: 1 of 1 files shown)


MatthewZMD:
Haven't seen you around in the past couple days :)

> The implementation assumes app commands always use an underscore, I'm not sure this is the best way to go but that might suffice for now.

Originally the underscore is to differentiate Python variables from emacs variables

MatthewZMD:
> What do you think about additionally indicating the app name in the mode line? Maybe something like:
> 
> ```
> <buffer-name>[<app-name>]
> ```
> 
> Or maybe we could append the app name to the mode-name:
> 
> ```
> EAF: <app-name>
> ```

This sounds like a great idea! I prefer the latter option `EAF: <app-name>`, `buffer-name` can get crowded sometimes ;)

PR #176, opened 2019-12-22
pdf-viewer: add cursor select text mode.

luhuaei:
Default double enter select mode, move mouse will highlight the select area. Current only has copy action, select mode will auto cancel on action finish.


luhuaei:
repeat select will cause EAF segmentation error. close pull requests

PR #177, opened 2019-12-25
Restore session for pdf bookmarks

clemera:
If eaf has not initilized this works but it shows following message:

    dbus-call-method: D-Bus error: "The name com.lazycat.eaf was not provided by any .service files"

on first invokation. Don't know how to fix that.

### DIFF
--- a/core/buffer.py
+++ b/core/buffer.py
@@ -260,6 +260,9 @@ class Buffer(QGraphicsScene):
     def execute_function(self, function_name):
         getattr(self, function_name)()
 
+    def execute_function_with_arg(self, function_name, arg):
+        getattr(self, function_name)(arg)
+
     def call_function(self, function_name):
         return getattr(self, function_name)()
 
--- a/eaf.py
+++ b/eaf.py
@@ -255,6 +255,14 @@ class EAF(dbus.service.Object):
             except AttributeError:
                 self.message_to_emacs("Cannot execute function: " + function_name)
 
+    @dbus.service.method(EAF_DBUS_NAME, in_signature="sss", out_signature="")
+    def execute_function_with_arg(self, buffer_id, function_name, arg):
+        if buffer_id in self.buffer_dict:
+            try:
+                self.buffer_dict[buffer_id].execute_function_with_arg(function_name, arg)
+            except AttributeError:
+                self.message_to_emacs("Cannot execute function: " + function_name + "with arg: " + arg)
+
     @dbus.service.method(EAF_DBUS_NAME, in_signature="ss", out_signature="s")
     def call_function(self, buffer_id, function_name):
         if buffer_id in self.buffer_dict:
--- a/eaf.el
+++ b/eaf.el
@@ -376,6 +376,8 @@ For now only EAF browser app is supported."
 (defun eaf--pdf-viewer-bookmark ()
   `((handler . eaf--bookmark-restore)
     (eaf-app . "pdf-viewer")
+    (session . ,(eaf-call "call_function"
+                          eaf--buffer-id "save_session_data"))
     (defaults . ,(list eaf--bookmark-title))
     (filename . ,(eaf-get-path-or-url))))
 
@@ -385,7 +387,11 @@ For now only EAF browser app is supported."
     (cond ((equal app "browser")
            (eaf-open-browser (cdr (assq 'filename bookmark))))
           ((equal app "pdf-viewer")
-           (eaf-open (cdr (assq 'filename bookmark)))))))
+           (eaf-open (cdr (assq 'filename bookmark)))
+           (eaf-call "execute_function_with_arg"
+                     eaf--buffer-id
+                     "restore_session_data"
+                     (cdr (assq 'session bookmark)))))))
 
 (defun eaf-open-bookmark ()
   "Command to open or create EAF bookmarks with completion."


core/buffer.py
```diff
@@ -260,6 +260,9 @@ def update_with_data(self, update_data):
     def execute_function(self, function_name):
         getattr(self, function_name)()
 
+    def execute_function_with_arg(self, function_name, arg):
+        getattr(self, function_name)(arg)
+
```
MatthewZMD: Instead of creating an additional function to execute with args, maybe we can reuse `execute_function` like this?
```
>>> class Class:
...     def foo(self, fname, *args):
...             getattr(self, fname)(*args)
```

eaf.py
```diff
@@ -255,6 +255,14 @@ def execute_function(self, buffer_id, function_name):
             except AttributeError:
                 self.message_to_emacs("Cannot execute function: " + function_name)
 
+    @dbus.service.method(EAF_DBUS_NAME, in_signature="sss", out_signature="")
+    def execute_function_with_arg(self, buffer_id, function_name, arg):
+        if buffer_id in self.buffer_dict:
+            try:
+                self.buffer_dict[buffer_id].execute_function_with_arg(function_name, arg)
+            except AttributeError:
+                self.message_to_emacs("Cannot execute function: " + function_name + "with arg: " + arg)
```
MatthewZMD: Same as above, I think we should try to reuse `execute_function`. 

In this case, make the argument `args` defaulted to an empty string, and with a similar hack we did in `store_emacs_var` function:
```
@dbus.service.method(EAF_DBUS_NAME, in_signature="s", out_signature="")
    def store_emacs_var(self, var_dict_string):
        for var_pair in var_dict_string.split(":"):
            (var_name, var_value) = var_pair.split(",")
```
So that we can call Python function with arbitrary number of arguments :)

MatthewZMD:
> If eaf has not initilized this works but it shows following message:
> 
> ```
> dbus-call-method: D-Bus error: "The name com.lazycat.eaf was not provided by any .service files"
> ```
> 
> on first invokation. Don't know how to fix that.

This should fix the error

```
-           (eaf-open (cdr (assq 'filename bookmark)))))))
+           (eaf-open (cdr (assq 'filename bookmark)))
+           (when (derived-mode-p 'eaf-mode)
+             (eaf-call "execute_function_with_arg"
+                       eaf--buffer-id
+                       "restore_session_data"
```

manateelazycat:
Default, EAF will restore session when PDF file opened, why need add session operation in bookmark function?

MatthewZMD:
Yes, EAF PDF's current behavior is automatically updating session data upon exit, and automatically loading from session data upon opening a pdf file.

MatthewZMD:
There is a problem with this approach: since this will load from the session file when opening pdf from the bookmark, if a user is many pages ahead of the last time he closes Emacs or kills the eaf-pdf buffer, and he goes to edit other files. Then if he wants to continue reading pdf again, but tries to open the pdf from the bookmark - which many people do - will cause the page jumps to the last time he saves session.

core/buffer.py
clemera: I think then we have to call all `execute_function` always with empty string as argument if it doesn't take any, but I haven't tested it. I will later look into it.

eaf.py
clemera: How would the signature have to be declared then:

    @dbus.service.method(EAF_DBUS_NAME, in_signature="s", out_signature="")

I thought to go with one arg and use a string that gets splitted, because I don't know if there is another way to declare multiple args.

clemera:
> Default, EAF will restore session when PDF file opened, why need add session operation in bookmark function?

The sessions gets stored automatically so if you reopen a pdf it automatically jumps to to last session. When I make a bookmark I want to remember a special location/state and be able to visit it at a later time. For example say you want to save the location of a code example given in a pdf and then you continue reading. When you reopen it jumps to the last location where you stopped reading but with the bookmark you can quickly jump to the code example you looked at.

clemera:
> Then if he wants to continue reading pdf again, but tries to open the pdf from the bookmark - which many people do - will cause the page jumps to the last time he saves session.

That is intentionally from my side ;) But I understand the problem, if you are using bookmarks this way. I think bookmarks should only be used to save the exact state (session) you are in and opening the last session should be done via other means.

MatthewZMD:
> go with one arg and use a string that gets splitted

Yes, this is what I meant, I looked into D-Bus specification but cannot find anywhere about arbitrary parameters.

core/buffer.py
MatthewZMD: > I think then we have to call all execute_function always with empty string as argument if it doesn't take any

Yes, `eaf-call` has to call `execute_function` with empty string.

MatthewZMD: > I think then we have to call all execute_function always with empty string as argument if it doesn't take any

Yes, `eaf-call` has to call `execute_function` with empty string.

MatthewZMD:
> When I make a bookmark I want to remember a special location/state and be able to visit it at a later time. For example say you want to save the
> location of a code example given in a pdf and then you continue reading. When you reopen it jumps to the last location where you stopped reading but with the bookmark you can quickly jump to the code example you looked at.

I discussed with Andy, and right, this is the only use case we can think of.

However, as I mentioned earilier, there is that problem if handling sessions with bookmarks.

> with the bookmark you can quickly jump to the code example you looked at.

If you have a long list of bookmarks, it would somtimes be actually more troublesome to find the right bookmark and jump back to where you left off.


Andy suggested that if we really want to handle such use case, it would be better to create a new functionality that temporarily remembers a position instead of messing with the session file.

clemera:
> instead of messing with the session file.

I was not aware of that, `save_session_data` only returns the session data and does not impact the session file I thought.

clemera:
> This should fix the error, adding a (when (derived-mode-p 'eaf-mode) check.

Hm, the problem seems to be the process is not started yet when the `eaf-call` gets executed. We would need a way to execute things after the process started. When I use `(run-at-time 1 ...)` it works. Maybe we should add a sentinel that runs on after the process started to execute code that might be needed?

manateelazycat:
There are two types of behavior:

1. The bookmark save page information, enter the path to open each PDF file, so session file is perfect to jump last reading postion, and not conflict with bookmark.
2. Like me, I organize PDF files to a directory structure inside, and too lazy to look where PDF file is, I can open the PDF file and continue to reading last position, I use bookmark to open file, if we save page information in bookmark, it's too painful to me.

I'm not saying that this patch is wrong, but everyone is not the same bookmarks habits, I also need to think about, to see if there a way out?

clemera:
Yes, I haven't thought about that problem before. My use case is for example having multiple bookmarks for a book to jump to specific pages. One way I can think of is not adding this patch and just create a `eaf-restore-bookmark-alist` like we have for creating the bookmark. Then users can configure it the way they want making the old behavior the default.

clemera:
Should I work on refactoring the bookmark code to support that use case instead? Then nothing of current behavior changes and the new behavior is optional.

manateelazycat:
> Should I work on refactoring the bookmark code to support that use case instead? Then nothing of current behavior changes and the new behavior is optional.

yes of course, i will sleep soon , see you

clemera:
I will look into it the next view days, will reopen another PR then

MatthewZMD:
> Maybe we should add a sentinel that runs on after the process started to execute code that might be needed?

Right, we should do that

MatthewZMD:
> For example say you want to save the location of a code example given in a pdf and then you continue reading. When you reopen it jumps to the last location where you stopped reading but with the bookmark you can quickly jump to the code example you looked at.

I *just* realized we already have functions `remember_current_position` mapped to `[` for remembering current position and `remember_jump` mapped to `]` for jumping to that remembered position. 

These are very misleading function names that I totally forgot the feature even existed... I'm gonna change the names now, but doesn't it fulfill this example's need already?

clemera:
Those are useful indeed and I haven't tried them before but you can only remember one position and it isn't remembered across sessions.  

Another thing is you can't give it a name like bookmarks. I think it's great  as a temporary thing to jump between two places in a pdf you are currently reading.

PR #180, opened 2019-12-29
Make window-combination-resize setq-local

MatthewZMD:
@clemera I saw your FIXME regarding `(setq window-combination-resize t)`. Would `setq-local` fix this problem?

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -7,7 +7,7 @@
 ;; Copyright (C) 2018, Andy Stewart, all rights reserved.
 ;; Created: 2018-06-15 14:10:12
 ;; Version: 0.5
-;; Last-Updated: Sun Dec 29 00:55:29 2019 (-0500)
+;; Last-Updated: Sun Dec 29 01:06:14 2019 (-0500)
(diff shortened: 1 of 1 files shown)


clemera:
I'm not sure what the reason is to set this setting. Making it buffer local means it only applies when splitting windows from the eaf buffer. This does not change the global user setting so this would fix the problem of changing the potentially unwanted splitting behavior. But maybe it can be removed completely?

MatthewZMD:
@manateelazycat

manateelazycat:
> I'm not sure what the reason is to set this setting. Making it buffer local means it only applies when splitting windows from the eaf buffer. This does not change the global user setting so this would fix the problem of changing the potentially unwanted splitting behavior. But maybe it can be removed completely?

The reason is that the very beginning, if not so set, the more points the smaller the window, the window is too small, EAF would not be able to use, such as the browser.

MatthewZMD:
Yes, so the problem right now is EAF changes global user setting. Shall we change it to `setq-local`?

PR #188, opened 2020-01-03
README update

tumashu:


manateelazycat:
Hi

Thanks for patch first.

README has include python3 note, and user should figure out how to install dependences of pymupdf.

README docs shouldn't include apt-get or pacman instruction.

tumashu:
ok

Issue #181, opened 2019-12-30
elisp: render HTML region


JulienMasson:
Is it possible to render HTML region in a buffer ?

The goal I have in mind is to replace the current html render of my MUA with *eaf*.
I've been able to export the html part of my mail and open it in an external buffer with *eaf*.
The rendering is perfect, it's really good thanks for your work.
But I was wondering if it's possible to render only the body (html) of my mail buffer.
Something similar to `shr-render-region`.
I don't know if it's possible with the XReparent technology ...

Thanks

MatthewZMD:
A possible solution I can think of right now is create a function that creates a html file at the `/tmp` directory, then eaf-open it

manateelazycat:
no its not possible to do that,eaf design to replace all buffer content,it can not mix eaf in emacs buffer region

MatthewZMD:
```
(defun eaf-render-region-as-html ()
  "Assuming marked region is HTML, create a tmp file and use EAF to render the region."
  (interactive)
  (when mark-active
    (let ((tmp-file "/tmp/eaf-render-region.html"))
      (write-region (region-beginning) (region-end) tmp-file)
      (eaf-open tmp-file)
      (delete-file tmp-file))))
```

Here is a function fulfilling your request. 

Dunno whether I should push it to upstream or not. @manateelazycat

JulienMasson:
@manateelazycat, okay yes I expected that behavior.

@MatthewZMD, yes I wrote similar function for my case:
```
(defun mail-open-external-browser (html)
  (interactive)
  (let* ((filename (concat (make-temp-name "mail-") ".html"))
	 (file (concat (temporary-file-directory) filename)))
    (with-temp-file file
      (insert html))
    (eaf-open file)))
```

For rendering an html region I was thinking about this kind of case:
```
(defun compose-html-mail (html)
  (compose-mail)
  (message-goto-body)
  (let ((beg (point-min)))
    (insert html)
    (eaf-render-region beg (point-max))))
```

MatthewZMD:
Are you using mu4e? I am thinking EAF should support an option to automatically render html emails when possible. Would you mind expanding your function on extracting `html` content of the email and either make a PR to here or to the Wiki, that would be best.

JulienMasson:
Euh not anymore I wrote my mua but I have used/hacked different mua like gnus, notmuch, mu4e. 

[...]

PR #193, opened 2020-01-06
Add new command eaf-open-mail

JulienMasson:
The goal of this command is to open any html mail in EAF browser.
Several Mail User Agent are supported in this patch:
- gnus
- mu4e
- notmuch

WARNING:
A temporary file is created to store html contents.
This file is NOT auto deleted once the EAF browser is closed.

Signed-off-by: Julien Masson <massonju.eseo@gmail.com>

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -342,6 +342,16 @@ Try not to modify this alist directly.  Use `eaf-setq' to modify instead."
   "The extension list of org previewer application."
   :type 'cons)
 
+(defcustom eaf-mua-get-html
+  '((gnus-user-agent    . eaf-gnus-get-html)
+    (mu4e-user-agent    . eaf-mu4e-get-html)
+    (notmuch-user-agent . eaf-notmuch-get-html))
+  "Association list with mail user agent as a KEY and a function as VALUE used
+to retrieve HTML part of a mail.
+
+The value of `mail-user-agent' has to be a KEY of `eaf-mua-get-html'."
+  :type 'alist)
+
 (defvar eaf-app-binding-alist
   '(("browser" . eaf-browser-keybinding)
     ("pdf-viewer" . eaf-pdf-viewer-keybinding)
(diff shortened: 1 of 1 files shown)


JulienMasson:
This patch is related to the discussion #181 

This is a PoC, please let me know if you prefer to handle this differently :)

manateelazycat:
We can add the interface to open html string in eaf browser，then we can avoid temp html file

eaf.el
```diff
@@ -939,6 +949,67 @@ of `eaf--buffer-app-name' inside the EAF buffer."
     (switch-to-buffer buf)
     (other-window +1)))
 
+(defun eaf--gnus-htmlp (part)
+  (when-let ((type (mm-handle-type part)))
+    (string= "text/html" (car type))))
+
+(defun eaf--nomtuch-htmlp (part)
```
MatthewZMD: I think you misspelled notmuch

MatthewZMD:
Thanks for the patch! I am in the office right now and cannot properly test it, but I think I've found a small typo, please see the review comment.

eaf.el
```diff
@@ -939,6 +949,67 @@ of `eaf--buffer-app-name' inside the EAF buffer."
...
+	      (body (plist-get msg :body))
+	      (parts (car body))
+	      (content (plist-get parts :content))
+	      (part (if (listp content)
+			(cl-find-if #'eaf--nomtuch-htmlp content)
+		      (when (eaf--nomtuch-htmlp parts)
+			parts))))
+    (notmuch-get-bodypart-text msg part notmuch-show-process-crypto)))
+
+(defun eaf-open-mail ()
```
MatthewZMD: Maybe name it something like `eaf-open-mail-as-html` ?

MatthewZMD:
> This file is NOT auto deleted once the EAF browser is closed.

I think you can simply delete the temporary file after calling `eaf-open`, the EAF buffer will not get deleted.

Apparently my `mail-user-agent` defaults to `message-user-agent`, I have to manually put `(setq mail-user-agent (mu4e-user-agent))` in my config to make this work. Is it possible to make it automatically identify which MUA currently in? Lile using `(derived-mode-p 'mu4e-view-mode)`

MatthewZMD:
> avoid temp html file

We could also do this small hack
```
(eaf-open file)
(run-at-time 1 nil (lambda () (delete-file file)))
```

manateelazycat:
> > avoid temp html file
> 
> We can also do this small hack, it won't hurt ;)
> 
> ```
> (eaf-open file)
> (run-at-time 1 nil (lambda () (delete-file file)))
> ```

It's best delete temp file when QtWebEngine load page finish.

MatthewZMD:
> when QtWebEngine load page finish.

Is it possible to track when does QtWebEngine finish loading?

manateelazycat:
> > when QtWebEngine load page finish.
> 
> Is it possible to track when does QtWebEngine finish loading?

In file core/browser.py: 
```
self.buffer_widget.loadFinished.connect(self.stop_progress)
```

manateelazycat:
There have two way:

1. Pass temp html file to eaf-open, then delete temp html file after QtWebEngine load finish.
2. Build a new interface by https://doc.qt.io/qt-5/qwebengineview.html#setHtml , then pass html string to EAF process directly.

I prefer temp html file since it's ugly that pass big data through DBus.

JulienMasson:
Another solution is to delete the temporary file when we receive `eaf--update-buffer-title` signal ?
That's a little bit dirty since it's not really the purpose of this function ... 

Maybe another dbus signal like `eaf--buffer-created` can be used/created for this kind of use case ?

eaf.el
```diff
@@ -939,6 +949,67 @@ of `eaf--buffer-app-name' inside the EAF buffer."
...
+In that way the corresponding function will be called to retrieve the HTML part of
+the current mail."
+  (interactive)
+  (when-let* ((get-html (eaf--get-html-func))
+	      (html (funcall get-html))
+	      (filename (concat (make-temp-name "eaf-mail-") ".html"))
+	      (file (concat (temporary-file-directory) filename)))
+    (with-temp-file file
+      (insert html))
+    (eaf-open file)))
```
manateelazycat: I have push new patch https://github.com/manateelazycat/emacs-application-framework/commit/8d9d78f2adf2521b18e41137cae62127b4c1219e

Change (eaf-open file) to 
```Elisp
(eaf-open file "browser" "temp_html_file")
```

Then EAF will read content of html to render, and delete temp file automatically

manateelazycat:
> Another solution is to delete the temporary file when we receive `eaf--update-buffer-title` signal ?
> That's a little bit dirty since it's not really the purpose of this function ...
> 
> Maybe another dbus signal like `eaf--buffer-created` can be used/created for this kind of use case ?

No, it's hacking way, use 
```
(eaf-open file "browser" "temp_html_file")
```

JulienMasson:
Thanks @manateelazycat I updated my commit with your last change.

manateelazycat:
thanks man

MatthewZMD:
Great, but I don't think my precious comment has been addressed yet 

> Apparently my `mail-user-agent` defaults to `message-user-agent`, I have to manually put `(setq mail-user-agent (mu4e-user-agent))` in my config to make this work.
> 
> Is it possible to make it automatically identify which MUA currently in? Like using `(derived-mode-p 'mu4e-view-mode)`

MatthewZMD:
Also, strangely I am receiving errors that I haven't receive yesterday

```
Debugger entered--Lisp error: (void-variable eaf-mu4e-get-html)
  eval(eaf-mu4e-get-html)
  (and t (eval (assoc-default mail-user-agent eaf-mua-get-html)))
  (let* ((get-html (and t (eval (assoc-default mail-user-agent eaf-mua-get-html))))) (if get-html get-html (error "[EAF] Mail User Agent \"%s\" not supported! See all ..." mail-user-agent)))
  eaf--get-html-func()
  (and t (eaf--get-html-func))
  (let* ((html (and t (eaf--get-html-func))) (file (and html (concat (temporary-file-directory) (make-temp-name "eaf-mail-") ".html")))) (if file (progn (let ((temp-file file) (temp-buffer (generate-new-buffer " *temp file*"))) (unwind-protect (prog1 (save-current-buffer (set-buffer temp-buffer) (insert html)) (save-current-buffer (set-buffer temp-buffer) (write-region nil nil temp-file nil 0))) (and (buffer-name temp-buffer) (kill-buffer temp-buffer)))) (eaf-open file "browser" "temp_html_file"))))
```

Looks like `eaf-mu4e-get-html` is been evaluated as a variable instead of a function

MatthewZMD:
> Also, strangely I am receiving errors that I haven't receive yesterday
> 
> ```
> Debugger entered--Lisp error: (wrong-type-argument char-or-string-p eaf-mu4e-get-html)
>   insert(eaf-mu4e-get-html)
>   (save-current-buffer (set-buffer temp-buffer) (insert html))
>   (prog1 (save-current-buffer (set-buffer temp-buffer) (insert html)) (save-current-buffer (set-buffer temp-buffer) (write-region nil nil temp-file nil 0)))
> ```
> 
> Looks like `eaf-mu4e-get-html` didn't get evaluated as a function

Nevermind, this error is introduced from b6f54c8, I can fix it

JulienMasson:
> Instead of letting the user customize mail-user-agent themselves, we should automatically detect the current MUA

@MatthewZMD yes sorry I forgot about this one.
Well it adds more complexity because you can have different `major-mode` for a single MUA.
For example in notmuch the major-mode of the buffer can be `notmuch-show-mode` or `notmuch-tree-mode`.
Same case for gnus if your current buffer is the "Article" or the "Summary", the major mode is different.
That's why I made the choice to only check `mail-user-agent`.
When you require notmuch, `mail-user-agent` is set automatically:
https://github.com/notmuch/notmuch/blob/master/emacs/notmuch.el#L1133
But not all the MUA do this ...

So as you want, we change the logic and check the major mode if you prefer

MatthewZMD:
Sounds fair. Another possible solution to reduce complexity is to check if the major-mode start with gnus, notmuch mu4e prefixes

PR #212, opened 2020-02-01
allow user set where store cache files

ctrl2wei:


MatthewZMD:
Maybe name it something other than cache files, because it is not just cache.

ctrl2wei:
> Maybe name it something other than cache files, because it is not just cache.

I have no idea how to name it. I just don't want it in my configuration directory.

MatthewZMD:
No worries. Please change it to eaf-config-location

MatthewZMD:
LGTM, thanks for the patch!

PR #216, opened 2020-02-02
Allow user define search engine and switch it when call eaf-search-it…

ctrl2wei:
… interactively

BREAKING CHANGE: search engine name become a string.

Use eaf-browser-search-engines store all search engine,
with a prefix arg, eaf-search-it is able to choose a
search engine to search.


MatthewZMD:
LGTM, thank you for the patch!

PR #217, opened 2020-02-02
fix/rss: use link when guid is not available

FirstLoveLife:
Some rss doesn't provide a id/uuid, so they cannot provide
a attr named "id", which will result of a coredump.

"link" is a good alternative when "id/guid"
is not provided

How to reproduce coredump? just add
"http://lambda-the-ultimate.org/rss.xml" to rss-reader


manateelazycat:
Thanks for patch ;)

  -- powered by EAF browser.

PR #220, opened 2020-02-03
fix commit 319aeda, replace defvar with defcustom

ctrl2wei:


MatthewZMD:
thanks for the patch!

PR #221, opened 2020-02-04
Pdf-viewer: add highlight annot on select mode.

luhuaei:
On select mode press h will highlight current select text and save as highlight annot on PDF file.


MatthewZMD:
Thanks for the patch :smiley:

PR #223, opened 2020-02-05
pdf-view: tell user how to select pdf text when copy_select copy fail

tumashu:


app/pdf-viewer/buffer.py
```diff
@@ -162,7 +162,7 @@ def copy_select(self):
             self.message_to_emacs.emit(content)
             self.buffer_widget.cleanup_select()
         else:
-            self.message_to_emacs.emit("Cannot copy, you haven't select anything!")
+            self.message_to_emacs.emit("Cannot copy, you should double click left mouse button then drag to select pdf text!")
```
MatthewZMD: you should double click your mouse and hover through the text on the PDF. Don't click and drag!

MatthewZMD: s/Don't/Do not

tumashu:
Fixed, but I think we hade better support two drag mode.

tumashu:
1. 2click + drag  
2. 2click+  hold button drag

PR #224, opened 2020-02-05
Fix: don't input Chinese or Unicode char.

luhuaei:
current could use browser input Chinese on EAF browser.


MatthewZMD:
感谢！这个很可以

PR #246, opened 2020-03-16
Add some terminal key

wi24rd:


MatthewZMD:
Thank you for the patch

PR #258, opened 2020-03-29
better detection of private codec support

wsw0108:


manateelazycat:
Thanks for patch

manateelazycat:
Hi

I just commit new patch https://github.com/manateelazycat/emacs-application-framework/commit/9c7ab52667f8e3e66280742cee340ce8ee8ff3ff

We can finish ldd test in python side, avoid get path and test at elisp side, make code much simpler.

Please let me know if you have any problem with new path.

Thanks again for patch!

wsw0108:
Thanks for your work.

I have create another PR #259 to fix some problem.

PR #262, opened 2020-03-30
Added Python version check

alaiyeshi025:
parameter "text" is added to subprocess.run since 3.7.


manateelazycat:
Thanks for great patch!

PR #274, opened 2020-04-12
Fix horizontal scrolling into a smooth movement

tobiasBora:
It fixes some issues with the horizontal scrolling (for some reasons, scrolling with the mouse vertically would reset the horizontal scroll). I also added a smooth horizontal scroll (using the same idea as the vertical scroll). It seems to work pretty well now when you scroll either horizontally or vertically.

It's still not perfect however when you scroll both at the same time (it's a bit choppy and tends to favorize one direction over the other), but it's much easier to follow now. I guess it's a "feature" of the event QEvent object, but in our case it makes diagonal scrolling choping.


manateelazycat:
Thanks for patch

PR #285, opened 2020-04-28
Prevent buffer name collision

AloisJanicek:
Prevent buffer name collision by demanding unique buffer name.

When having opened _WEB PAGE 1_ in EAF browser _BUFFER 1_, opening
_WEB PAGE 1_ from another EAF browser buffer results in error.

Passing `t` as a second argument to `rename-buffer` will suffix
duplicate buffer names with `<NUM>`.


MatthewZMD:
Great, thank for the patch!

AloisJanicek:
I am glad I could contribute. Just one character but it makes difference :-)

tumashu:
推敲

PR #306, opened 2020-06-02
Compatible support open pdfview: pdftools: links with EAF PDF Viewer

stardiviner:
This new feature is archived through enable option
`eaf-org-override-pdf-links' to override PDF file link types :follow
function. This can be restore to original :follow function automatically
when option is nil.


MatthewZMD:
赞👍🏻

PR #316, opened 2020-06-28
support emacs 27 tab-line-mode

wpchou:
tab-line-mode news: https://github.com/emacs-mirror/emacs/blob/master/etc/NEWS.27#L2755


MatthewZMD:
Nice

PR #322, opened 2020-07-02
Terminal: Fix error in connection establishment: net::ERR_CONNECTION_REFUSED

HollowMan6:
Successfully reproducing this bug:
image

When having this situation, you can still access the terminal by click the right button of the mouse, in the menu clicking `reload`, and then everything works fine.

So I believe this bug is caused by loading the web page too early , and I made a exception handling, when this happens, refresh the page automaticly after 1 second, and that bug never happens again.

### DIFF
--- a/app/terminal/index.html
+++ b/app/terminal/index.html
@@ -33,7 +33,12 @@
         <div id="xterm" class="xterm"/>
 
         <script>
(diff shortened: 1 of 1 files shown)


MatthewZMD:
赞，我回头测试一下，没问题就merge

MatthewZMD:
赞

PR #327, opened 2020-07-04
Mindmap: Enter color from the keyboard to change node color.

HollowMan6:
This is the new design for #326 according to the feedback.

I think the new design gives users mush flexibility and it's the best than ever.

Users can input any HTML color codes instead of using a color list.

When you want to change node background color, just press 'c', and then input any format of the HTML color codes.

When you want to change node text color, just press 'C', and then input any format of the HTML color codes.

| Entering Format| Before| After|
| :--------: | :--------: | :----: |
|Color Name|  ||
|RGB Color Code|  ||
|Hex Color Code|  ||


manateelazycat:
超级赞！

PR #341, opened 2020-07-17
Browser: Autofill form by storing form data in a database

HollowMan6:
The database is located at `config_dir/browser/password.db`
### Database Structure
image
`password` collects the input that has type "password".
`form_data` collects all the input fields in the current page except type "password" and "hidden".

## Record form
Under my configuration, before you are ready to login, you can press `C-d` to record form data into database:
image
You can add multiple accounts into database.
image
If there already exists a record in the database that has the same `form_data`, it will only update the password.
It can also support record form without password.

## Autofill
If you have enabled the autofill, it will fill the form starting from the smallest `id` in database. Under my configuration, press `C-t` to toggle autofill and switch between different accouts.
Signed-off-by: Hollow Man <hollowman186@vip.qq.com>


MatthewZMD:
赞，我回头测一下

PR #347, opened 2020-07-24
customize pdf viewer's default zoom.

zbelial:


manateelazycat:
only eaf-pdf-default-zoom one option is enough
if its not 0, use default zoom

app/pdf-viewer/buffer.py
```diff
@@ -204,6 +204,9 @@ def __init__(self, url, config_dir, background_color, buffer_id, emacs_var_dict)
         # Init scale and scale mode.
         self.scale = 1.0
         self.read_mode = "fit_to_width"
+        if (self.emacs_var_dict["eaf-pdf-read-mode"] == "fit_to_customize"):
+            self.read_mode = "fit_to_customize"
+            self.scale = float(self.emacs_var_dict["eaf-pdf-default-zoom"])
```
MatthewZMD, line 209: Like @manateelazycat  said, an easier logic: if `eaf-pdf-default-zoom` is not `1`, set `read_mode` to `fit_to_customize` and set `self.scale` to the `eaf-pdf-default-zoom`.

zbelial:
Thanks. 
I have changed the code according to your advice.  Please review again.

manateelazycat:
thanks for patch

Issue #314, opened 2020-06-28
eaf 打开 epub 文件出错


loyalpartner:
```
RuntimeError: not a PDF
EAF process starting...
Session restored:  3fb2-ae25-6a7b-7c6c-4301-ec25-ddf6 app.pdf-viewer.buffer /home/lee/.emacs.d/eaf/session.json
mupdf: not a PDF
Traceback (most recent call last):
  File "/home/lee/.emacs.d/.local/straight/repos/emacs-application-framework/app/pdf-viewer/buffer.py", line 382, in paintEvent
    qpixmap = self.get_page_pixmap(index, self.scale)
  File "/home/lee/.emacs.d/.local/straight/repos/emacs-application-framework/app/pdf-viewer/buffer.py", line 310, in get_page_pixmap
    self.char_dict[index] = self.get_page_char_rect_list(index)
  File "/home/lee/.emacs.d/.local/straight/repos/emacs-application-framework/app/pdf-viewer/buffer.py", line 742, in get_page_char_rect_list
    page_rawdict = self.document[page_index].getText("rawdict")
  File "/usr/lib/python3.8/site-packages/fitz/utils.py", line 463, in getText
    tp = page.getTextPage(flags)  # TextPage with or without images
  File "/usr/lib/python3.8/site-packages/fitz/fitz.py", line 3608, in getTextPage
    self.setRotation(0)
  File "/usr/lib/python3.8/site-packages/fitz/fitz.py", line 4140, in setRotation
    return _fitz.Page_setRotation(self, rot)
RuntimeError: not a PDF
```

MatthewZMD:
@manateelazycat EAF现在还支持epub吗？我不用epub所以还没试过

manateelazycat:
Yes, mupdf support epub book.

Can you open same book in other epub reader?

loyalpartner:
@manateelazycat  我用fbreader是可以打开的，前面eaf打开epub没问题，昨天升级了下 eaf, 就遇到这个问题了

MatthewZMD:
你git checkout一下之前的commit，看一看是什么时候引入的问题

loyalpartner:
https://github.com/pymupdf/PyMuPDF/issues/540
是pymupdf的锅

loyalpartner:
https://github.com/manateelazycat/emacs-application-framework/blob/83a25b83c4fc831a73a16dcdb617446a09c29507/app/pdf-viewer/buffer.py#L331
[...]

PR #356, opened 2020-07-31
PDF-Viewer: Add restriction to PDF files only #314

HollowMan6:
> https://github.com/manateelazycat/emacs-application-framework/blob/83a25b83c4fc831a73a16dcdb617446a09c29507/app/pdf-viewer/buffer.py#L331
> 
> 

image

Since pymupdf doesn't support changing `epub` file's background currently, I add restriction so that `page.drawRect` only executes in the pdf files to solve #314 .

Signed-off-by: Hollow Man <hollowman186@vip.qq.com>


manateelazycat:
python有获取扩展名的api，不要用split方法

HollowMan6:
> python有获取扩展名的api，不要用split方法

OK，已经更新

PR #394, opened 2020-09-12
add total_page

zbelial:


app/pdf-viewer/buffer.py
```diff
@@ -147,6 +147,9 @@ def copy_select(self):
         else:
             self.message_to_emacs.emit("Cannot copy, you should double click your mouse and hover through the text on the PDF. Don't click and drag!")
 
+    def total_page(self):
```
MatthewZMD, line 150: Can you rename the function to page_total_number?

zbelial: No problem.

PR #412, opened 2020-10-19
improve pdf-viewer render quality

zbelial:
the pixmap and rectangle size should be the same, otherwise the pixmap will be scaled, leading to poor render quality. 
Find more information [here](https://doc.qt.io/qt-5/qpainter.html#drawPixmap-9).


zbelial:
pixmap的尺寸要跟实际绘图区域的尺寸完全一样，否则Qt就会缩放一下，导致看起来模模糊糊的。

MatthewZMD:
赞👍🏻

manateelazycat:
awesome

PR #420, opened 2020-10-22
Support eaf-browser-font-family and eaf-terminal-font-family

mackong:
Support eaf-browser-font-family and eaf-terminal-font-family, ex.
```
(eaf-setq eaf-browser-font-family "DejaVu Sans")
(eaf-setq eaf-terminal-font-family "Fira Code")
```


manateelazycat:
First, thanks for patch.

I think this PR need adjust:
1. Use "" is not empty string as if condition
2. Please remove merge commit

Thanks.

mackong:
1. Sorry, which *if* do you mean, js or python? I have tested, both is ok.
2. merge commit is removed.

app/terminal/index.html
```diff
@@ -44,6 +44,9 @@
           cursorBlink: true,
           theme: theme
       });
+      if ("%6") {
```
MatthewZMD, line 47: Use "" is not empty string

MatthewZMD (changes requested):

app/terminal/index.html
mackong: like ```(eaf-setq eaf-terminal-font-family "")```?

MatthewZMD: yes

mackong: image

After ```(eaf-setq eaf-terminal-font-family "")```, the rendered index.html is like this, and ```if ("")``` is no problem, so what i miss?

manateelazycat: I see, patch is merged, thanks for PR ;)

PR #425, opened 2020-10-25
fix(mindmap): make new node in edit mode

mackong:
Before, the *RET* key is send directly to jsmind, and *handle_addbrother* in jsmind is called, not our *add_brother_node* function,
which make *add brother* correct, but *add sub*, *add middle* is not.
Now we make *RET*, *TAB*, *i* call our functions and add *begin_edit* in our functions.

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -283,6 +283,7 @@ It must defined at `eaf-browser-search-engines'."
     (eaf-terminal-font-family . "")
     (eaf-mindmap-dark-mode . "follow")
     (eaf-mindmap-save-path . "~/Documents")
+    (eaf-mindmap-edit-mode . "false")
     (eaf-marker-letters . "ASDFHJKLWEOPCNM")
     (eaf-emacs-theme-mode . "")
     (eaf-emacs-theme-background-color . "")
--- a/app/mindmap/buffer.py
+++ b/app/mindmap/buffer.py
@@ -45,7 +45,11 @@ class AppBuffer(BrowserBuffer):
 
         self.cut_node_id = None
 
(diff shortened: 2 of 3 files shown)


manateelazycat:
I dislike node default use edit mode, because I think Emacs's minibuffer edit text much better than edit node directly.

Anyway, I think we can include an option to control whether enable edit mode when create node.

mackong:
*eaf-mindmap-edit-mode* option added.

BTW, edit in minibuffer is very convinience, havn't notice before, maybe the PR is useless, 😄

eaf.el
```diff
@@ -120,7 +120,7 @@
     (define-key map (kbd "M-[") #'eaf-share-path-or-url)
     (define-key map (vector 'remap #'keyboard-quit) #'eaf-keyboard-quit)
     (define-key map (vector 'remap #'self-insert-command) #'eaf-send-key)
-    (dolist (single-key '("RET" "DEL" "TAB" "SPC" "<backtab>" "<home>" "<end>" "<left>" "<right>" "<up>" "<down>" "<prior>" "<next>" "<delete>" "<backspace>" "<return>"))
+    (dolist (single-key '("RET" "DEL" "TAB" "SPC" "<backtab>" "<home>" "<end>" "<left>" "<right>" "<up>" "<down>" "<prior>" "<next>" "<delete>" "<backspace>"))
```
manateelazycat, line 123: Why remove <return> here?

manateelazycat (approved):

eaf.el
mackong: Sorry, removed when debuging. 
Reverted.

manateelazycat:
Thanks man. ;)

PR #433, opened 2020-10-29
fix(terminal): fix unscrollable and text search

mackong:
Make interactive methods in AppBuffer don't overrided by BrowserView with the same name.


MatthewZMD:
赞👍🏻

Issue #453, opened 2020-11-30
Forcing color inversion for PDFs with pages that are entirely images


Jdogzz:
**Is your feature request related to a problem? Please describe.**
I have found the PDF color inversion to work perfectly for example PDFs which consist of text with images throughout. However, for PDFs which are full page images, the color inversion does not activate and leaves the document rendered as it originally appears. To the best of my knowledge, the color inversion keybinding `i` has no effect on documents that fit this criteria (I seem to presently be suffering from #237 but I have been able to successfully toggle color inversion on other documents which consist of text with some images, so I think I would have been able to do the same with the documents that are full page images). I have gone through the list of keybindings and see no alternative color inversion which would force documents like this to be inverted. 

**Describe the solution you'd like**
I would like to have a key binding which forcefully applies the color inversion irrespective of regions classified as images.

**Describe alternatives you've considered**
Another possible acceptable solution is to add a third toggle when using `i` which cycles through: dark (leave images alone), dark (including images), original color.

manateelazycat:
@HollowMan6 can you look this issue?

MatthewZMD:
@HollowMan6 Can you add a variable that exclude #308 changes?

HollowMan6:
> @HollowMan6 Can you add a variable that exclude #308 changes?

Ok, but I'm busy with exams this week, and I'll look into this at the weekends when I have time.

HollowMan6:
Hi @Jdogzz , I've made the patch #457 adopting your alternative solution cuz I think it's a better way. Please feel free if you have any other suggestions. Thanks!

PR #457, opened 2020-12-01
pdf-viewer: Opimize toggle dark mode logic for #453

HollowMan6:
Adopted alternative solution in #453 by @Jdogzz 

Use `i` to toggle among `dark mode excluding images` -> `dark mode including images` -> `normal mode` -> ...

In addition, you can configure the emacs var dict to change the default behavior. If emacs var dict `eaf-pdf-dark-exclude-image` is true, the default dark mode will exclude images, otherwise it will include images.

Signed-off-by: Hollow Man <hollowman@hollowman.ml>

### DIFF
--- a/app/pdf-viewer/buffer.py
+++ b/app/pdf-viewer/buffer.py
@@ -243,6 +243,11 @@ class PdfViewerWidget(QWidget):
              self.emacs_var_dict["eaf-emacs-theme-mode"] == "dark")):
             self.inverted_mode = True
 
+        # Inverted mode exclude image.
+        self.inverted_mode_exclude_image = True
+        if (self.emacs_var_dict["eaf-pdf-dark-exclude-image"] == "false"):
+            self.inverted_mode_exclude_image = False
+
         # mark link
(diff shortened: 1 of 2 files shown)


app/pdf-viewer/buffer.py
```diff
@@ -243,6 +243,11 @@ def __init__(self, url, config_dir, background_color, buffer_id, emacs_var_dict)
              self.emacs_var_dict["eaf-emacs-theme-mode"] == "dark")):
             self.inverted_mode = True
 
+        # Inverted mode exclude image.
+        self.inverted_mode_exclude_image = True
+        if (self.emacs_var_dict["eaf-pdf-dark-exclude-image"] == "false"):
+            self.inverted_mode_exclude_image = False
```
manateelazycat, line 249: self.inverted_mode_exclude_image = self.emacs_var_dict["eaf-pdf-dark-exclue-image"] == "true" 

Is better way.

Anyway I have fix this problem in newest version.

PR #476, opened 2020-12-17
Fixes and Additions

Aeres-u99:
+ Added few dependencies of grip 
+ Fixed my mistakes
+ Tried updating the Zh-Cn README to the new additions 
---
Note:
Grip didnt work like a charm for me, apparently it was broken in the system and needed flask and Markdown. 
Could it be that arch doesnt packages grip with flask? Dont know. But pip installing flask and Markdown fixes it. 

Also I used google translate to translate the same in chinese, hopefully it should be error free. But please do check and confirm

Suggestion: 
We will need some sort of debugging (in emacs buffer using .message_to_emacs.emit(""), so that problems could be easily identified. Will look into it, if you feel its needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,14 +65,15 @@ You need install pymupdf from pip to avoid undefined symbol ```jbig2_page_out```
 ```
 sudo pacman -R python-pymupdf
 pip uninstall fitz
-pip install pymupdf
+# Grip has dependency of flask and Markdown
+pip install pymupdf grip flask Markdown
 ```
 
 2. For installing EAF, there are two routes:
 Please make sure to modify `~/emacs.d/site-lisp` as per your Emacs setup.
 - *Git Clone*:
 ```Bash
-git clone https://github.com/manateelazycat/emacs-application-framework.git --depth=1 ~/.emacs.d/site-lisp/
+git clone https://github.com/manateelazycat/emacs-application-framework.git --depth=1 ~/.emacs.d/site-lisp/emacs-application-framework/
 ```
 
 - *Quick Install*:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -85,6 +86,8 @@ Use *Quick Install* option if you want to quickly install EAF source code and tr
 3. From here on, you can either Add the full path to the EAF installation directory to your Emacs ```load-path```, then add the following to `init.el`:
 
 ```Elisp
+;; For debugging uncomment Following
+;; (message "Loading [EAF] ...")
```
MatthewZMD, line 90: This is very unnecessary

README.md
```diff
@@ -209,6 +212,12 @@ If you're sure Python dependences are installed correctly, please create an issu
 If you use EAF Markdown Previewer, to get consistent previewing, you need to access [Github Personal access token site](https://github.com/settings/tokens/new?scopes=), fill something in "Token description" and click button "Generate token" to get your personal token. Then set the token:
 
 ```Elisp
+;; for security reasons (if you add files to git) you can create secret.el with 
+;; following parameters and then load it (instead of setting directly inside init.el)
+;; put in the rigth path to replace, and add it to .gitignore
+;; (load "~/.emacs.d/secret.el")
+
+;; inside secret.el [or init.el if you are fine with that.]
```
MatthewZMD, line 220: I don't it's up to EAF to teach them this.

README.md
Aeres-u99: got it, please feel free to remove it. Or let me know if I should revert the commits and make new one with updated info.

README.md
Aeres-u99: true that's one thing, remove it if you feel its not okay, but we should try to promote good practices right?

README.md
MatthewZMD: You can simply add a new commit to your forked branch, it'll update it here automatically.

README.md
MatthewZMD: Yes, another option is to call customize-variable and save their value interactively, which will be saved to `customize-set-variables.el`, and if configured correctly, shouldn't be tracked by a VCS.

README.md
Aeres-u99: got it, I shall do the same with both then.

Aeres-u99: I shall close this one, and then open the one with required commits.

Aeres-u99:
Fixing the pr.

PR #477, opened 2020-12-17
Fixed as per conversation

Aeres-u99:
+ removed unnecessary verbosity 
+ removed Security thing I had added


MatthewZMD:
Thanks, but you don't need to close your previous PR and open a new one.

PR #483, opened 2020-12-22
WIP: Add EAF installation script

aancel:
This PR provides a script to ease dependency installations:
- Only required tools are installed via package managers
- Python dependencies are installed via pip using the requirements.txt file included

To test, this PR:
```shell
git clone -b feature/one-click-install https://github.com/aancel/emacs-application-framework
cd emacs-application-framework && ./install-eaf.sh
# Make sure everything goes fine or report issues in the comments here
```

Add the following configuration to you init file
```elisp
(use-package eaf
  ; Make sure to update the path according to your Emacs config location
  :load-path "~/.config/emacs/eaf/git/emacs-application-framework" 
)
```

Then launch:
```shell
# Make sure to update the path according to your Emacs config location
~/.config/emacs/eaf/bin/emacs-eaf.sh
```

Then make sure the different EAF applications are working.


manateelazycat:
@aancel

First, thanks for hard work!

Below is my suggestions:

1. Please just build script to install EAF dependencies, Emacs users have their own plug-in management system, such as, use-package, submodule and many more, we shouldn't force user git clone EAF at ~/.config/emacs/eaf/git/emacs-application-framework

2. Please use the Linux distribution's own package manager to install EAF dependence, take Arch as an example, python-pyqt5 is much stable than install from pip (many arch user report EAF browser crash that install pyqt5 from pip), I suggest Ubuntu script install package by apt. Yes I know it is easier to maintain through PIP, but the Linux distribution's own package manager is more stable, this is also the maintenance problem I said.

3. If user's Linux distribution is not apt or pacman, we should print package list let user hacking self, then send PR to us.

4. Please remove virtualenv stuff, reason same as point 2.

As I know, Ubuntu and Debian has it's own package version, such as pymupdf, Debian's package is much older than Ubuntu, distribution version should considered in installing scripts.

IMO, if we need to provide installation assist scripts, we need to seek best practical in each distribution, keep the user's stable use, not just our easy maintained (example use pip in Arch), this work is very difficult. ;)

aancel:
@manateelazycat 

Thanks for the feedback and the explanations, and sorry for the late answer !      
I've updated the script following your recommendations.     
Currently there still seems to be a separation between Ubuntu and Arch variants, notably in the naming of the package regarding python 3 (python3 vs python), which is a bit painful. Hopefully most of the packages will be able to be factorized, once the switch to python 3 will be definitive in further versions.     
Regarding the Arch version, I wasn't able to test it (I don't have access to my Manjaro PC right now), but it should be OK !

manateelazycat:
thanks for script，awesome

PR #490, opened 2020-12-31
terminal: trim preceding spaces of title

mackong:
if there are spaces between *:* and *\w* of $PS1, the terminal title will start with spaces, then the terminal buffer will not visible in emacs buffer list, and also you cannot select the buffer.


manateelazycat:
Thanks for commit. ;)

PR #496, opened 2021-01-07
Add dnf (Fedora) support to install.sh

dalanicolai:
I guess these must be all packages required for Fedora.
Only I am not completely sure about the equivalents for libdbus-1-3 and python-pyqt5, but I guess they are covered.
(Maybe it could be mentioned in the README that feedback for install.sh in Fedora is welcome. The comment can be removed when it is clear that all is fine)


MatthewZMD:
> (Maybe it could be mentioned in the README that feedback for install.sh in Fedora is welcome. The comment can be removed when it is clear that all is fine)

Thanks for the patch, could you propose a PR to update the README?

PR #497, opened 2021-01-07
Fix headers and eval eaf-enable-evil-intergration after eaf instead of evil

dalanicolai:
I am trying to create a layer for Spacemacs. However, the package would not install using a qualpa recipe because the headers had only 2 `;`'s instead of 3, ad [described here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html).

Also I tried to load `eaf-evil.el` :after `eaf`. However, with the `with-eval-after-load evil`, it would not load because clear_focus was not yet available. Changing it to ``with-eval-after-load eaf` fixed it.

### DIFF
--- a/eaf-evil.el
+++ b/eaf-evil.el
@@ -60,7 +60,7 @@
 
   (eaf-bind-key clear_focus "<escape>" eaf-browser-keybinding))
 
-(with-eval-after-load "evil"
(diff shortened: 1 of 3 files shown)


manateelazycat:
thanks for patch

PR #499, opened 2021-01-07
Update install instruction (add Fedora feedback request)

dalanicolai:
This would be okay?

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,13 +65,19 @@ EAF is an extensible framework, one can develop any Qt5 application and integrat
 git clone https://github.com/manateelazycat/emacs-application-framework.git --depth=1 ~/.emacs.d/site-lisp/emacs-application-framework/
 ```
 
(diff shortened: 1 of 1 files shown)


manateelazycat:
Thanks for patch! ;)

I think it's not need add those in README, user will send new PR if install-eaf.sh has bug in Fedora.

MatthewZMD:
That's true, thanks for the PR anyway!

dalanicolai:
> Thanks for patch! ;)
> 
> I think it's not need add those in README, user will send new PR if install-eaf.sh has bug in Fedora.

Well hope so. Sometimes it is not straightforward to find what goes wrong exactly, especially for new users. So I prefer to warn them, instead of let them wonder why things (that are expected to work) don't work. But indeed this would not be too difficult to trace, and indeed I guess most Emacs users are willing to report bugs. Well... it was just a proposal...
Thanks for your awesome work... EAF is really really nice!

PR #500, opened 2021-01-07
Make eaf-open-bookmark autoloadable.

dalanicolai:
This function is great for launching your favorite website which I often do at
first launch of my browser (e.g. visit github).

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -934,6 +934,7 @@ For now only EAF browser app is supported."
           ((equal app "pdf-viewer")
            (eaf-open (cdr (assq 'filename bookmark)))))))
 
+;;;###autoload
 (defun eaf-open-bookmark ()
   "Command to open or create EAF bookmarks with completion."
   (interactive)


core/browser.py
```diff
@@ -421,11 +421,12 @@ def refresh_page(self):
         ''' Refresh the page.'''
         self.reload()
 
+    @interactive(insert_or_do=True)
     def copy_text(self):
         ''' Copy selected text.'''
         self.triggerPageAction(self.web_page.Copy)
-        if self.buffer.caret_browsing_mode and self.buffer.caret_browsing_mark_activated:
-            self.buffer.caret_exit()
+        # if self.buffer.caret_browsing_mode and self.buffer.caret_browsing_mark_activated:
```
manateelazycat, line 428: why commet  those code?

manateelazycat (changes requested):

dalanicolai:
Ah sorry, definitely was not meant like this. This is a git/magit beginners mistake (well I use them both for quite some time now, but this never happened to me before). I just created this PR with just adding an  `;;;###autoloadable` comment added to the ` eaf-open-bookmark` function (just in the master branch as I am working on my own fork), subsequently I continued hacking.

I wanted to bind `copy_text` to `y` (vim-style), but then I could not type `y` into fields anymore. So I tried to create an `insert_or_copy_text` function. I pushed it to upstream because it was not working by adding that decorator, and I thought maybe it will work if the package gets reinstalled again (I am not sure how your package works exactly).

Anyway, I realize now that I messed up this PR. So it is a good lesson why I should make dedicated fix branches, especially for PR's.

dalanicolai:
Okay, I tried to fix it. So I did a hard reset on your (upstream) master branch head. Then I just added the ;;;###autoload comment, and I force pushed again. So I guess it looks fine now (or maybe not because the PR has multiple commits? I am still learning about git).

dalanicolai:
And I am still wondering how I can create an `insert_or_copy_text` function.

dalanicolai:
It would be great if in Doom/Spacemacs, the leader could just be on SPC (instead of C-SPC), while we could at the same time use SPC normally while inserting in text elements. I am trying now to not use the eaf-evil package and add `eaf-proxy-insert_or_focus_input` and `eaf-proxy-clear_focus` on the evil-insert-state entry and exit hooks. But I guess it would be great if it could be done in a cleaner way, only activate insert state when entering text in some text element.

loyalpartner:
> It would be great if in Doom/Spacemacs, the leader could just be on SPC (instead of C-SPC), while we could at the same time use SPC normally while inserting in text elements. I am trying now to not use the eaf-evil package and add `eaf-proxy-insert_or_focus_input` and `eaf-proxy-clear_focus` on the evil-insert-state entry and exit hooks. But I guess it would be great if it could be done in a cleaner way, only activate insert state when entering text in some text element.

The reason I don't set the leader key to spc is because there are other applications that need to use spc, if you want to use the leader key in the browser, you can refer to the following code
```
  (define-key key-translation-map (kbd "SPC")
    (lambda (prompt)
      (if (derived-mode-p 'eaf-mode)
          (pcase eaf--buffer-app-name
            ("browser" (if (eaf-call "call_function" eaf--buffer-id "is_focus")
                           (kbd "SPC")
                         (kbd "C-SPC")))
            ("pdf-viewer" (kbd "C-SPC"))
            ("image-viewer" (kbd "C-SPC"))
            (_  (kbd "SPC")))
        " ")))
```

loyalpartner:
> And I am still wondering how I can create an `insert_or_copy_text` function.
```
  (eaf-bind-key copy_text "y" eaf-browser-keybinding)
```

dalanicolai:
> The reason I don't set the leader key to spc is because there are other applications that need to use spc, if you want to use the leader key in the browser, you can refer to the following code

Thanks for the reply and the code snippet. But I do not understand your argument that other applications need to use SPC. SPC is indeed the main leader-key that is used for controlling Spacemacs/Doom. But also Spacemacs and Doom require SPC for entering text, that's why there is the insert state. I don't see any difference for any other application so that they require SPC other than for inserting text, it would be a choice of course. For example scrolling with SPC is a choice. I strongly prefer to use SPC like in Spacemacs and Doom and then scroll with `j/k`.

dalanicolai:
> > And I am still wondering how I can create an `insert_or_copy_text` function.
> 
> ```
>   (eaf-bind-key copy_text "y" eaf-browser-keybinding)
> ```

@loyalpartner Thanks for your reply. Indeed I had that. But then when I tried to insert/type text withing a webpage, e.g. google, I could not type `y` because it would call the `copy_text` command instead of self insert. I had the same problem when binding scroll_up/down, so that's when I discovered what `insert_or_scroll_up/down` was for...

loyalpartner:
> > > And I am still wondering how I can create an `insert_or_copy_text` function.
> > 
> > 
> > ```
> >   (eaf-bind-key copy_text "y" eaf-browser-keybinding)
> > ```
> 
> @loyalpartner Thanks for your reply. Indeed I had that. But then when I tried to insert/type text withing a webpage, e.g. google, I could not type `y` because it would call the `copy_text` command instead of self insert. I had the same problem when binding scroll_up/down, so that's when I discovered what `insert_or_scroll_up/down` was for...

```
    @interactive(insert_or_do=True)
    def copy_text(self):
        ''' Copy selected text.'''
        self.triggerPageAction(self.web_page.Copy)
        if self.buffer.caret_browsing_mode and self.buffer.caret_browsing_mark_activated:
            self.buffer.caret_exit()
```

loyalpartner:
> > The reason I don't set the leader key to spc is because there are other applications that need to use spc, if you want to use the leader key in the browser, you can refer to the following code
> 
> Thanks for the reply and the code snippet. But I do not understand your argument that other applications need to use SPC. SPC is indeed the main leader-key that is used for controlling Spacemacs/Doom. But also Spacemacs and Doom require SPC for entering text, that's why there is the insert state. I don't see any difference for any other application so that they require SPC other than for inserting text, it would be a choice of course. For example scrolling with SPC is a choice. I strongly prefer to use SPC like in Spacemacs and Doom and then scroll with `j/k`.

eaf has an app called eaf-terminal, if you bind `spc' to leader key, you can't type spc now.

dalanicolai:
Yes thanks, I tried that too... but that did not work either... that was exactly what I was doing and what I pushed accidentally into this PR.

dalanicolai:
> > > The reason I don't set the leader key to spc is because there are other applications that need to use spc, if you want to use the leader key in the browser, you can refer to the following code
> > 
> > 
> > Thanks for the reply and the code snippet. But I do not understand your argument that other applications need to use SPC. SPC is indeed the main leader-key that is used for controlling Spacemacs/Doom. But also Spacemacs and Doom require SPC for entering text, that's why there is the insert state. I don't see any difference for any other application so that they require SPC other than for inserting text, it would be a choice of course. For example scrolling with SPC is a choice. I strongly prefer to use SPC like in Spacemacs and Doom and then scroll with `j/k`.
> 
> eaf has an app called eaf-terminal, if you bind `spc' to leader key, you can't type spc now.

Yes that is why I think it would be nicer if there was still an insert state

dalanicolai:
> It would be great if in Doom/Spacemacs, the leader could just be on SPC (instead of C-SPC), while we could at the same time use SPC normally while inserting in text elements. I am trying now to not use the eaf-evil package and add `eaf-proxy-insert_or_focus_input` and `eaf-proxy-clear_focus` on the evil-insert-state entry and exit hooks. But I guess it would be great if it could be done in a cleaner way, only activate insert state when entering text in some text element.

@loyalpartner That is what I explained here. When binding SPC to leader, we can also not type in a webpage. However, when we do not use eaf-evil.el, then we can type SPC normall in a webpage from insert state, while in normal state it functions as the leader key. This is of course how all of Spacemacs (and probably Doom) works. I understand it might be a little more difficult to get the keybindings work like this in EAF, but I think your argument (and @manateelazycat ) do not hold.

Please explain if I do not understand you correctly...

dalanicolai:
Unfortunately when trying to use the terminal I get this (the buffer automatically get killed, but I was able to make a screenshot quickly)
Screenshot from 2021-01-08 11-50-33

dalanicolai:
The previous one was on Fedora... below is what I get on Ubuntu when trying to open `eaf-terminal`
Screenshot from 2021-01-08 12-24-02

loyalpartner:
> > It would be great if in Doom/Spacemacs, the leader could just be on SPC (instead of C-SPC), while we could at the same time use SPC normally while inserting in text elements. I am trying now to not use the eaf-evil package and add `eaf-proxy-insert_or_focus_input` and `eaf-proxy-clear_focus` on the evil-insert-state entry and exit hooks. But I guess it would be great if it could be done in a cleaner way, only activate insert state when entering text in some text element.
> 
> @loyalpartner That is what I explained here. When binding SPC to leader, we can also not type in a webpage. However, when we do not use eaf-evil.el, then we can type SPC normall in a webpage from insert state, while in normal state it functions as the leader key. This is of course how all of Spacemacs (and probably Doom) works. I understand it might be a little more difficult to get the keybindings work like this in EAF, but I think your argument (and @manateelazycat ) do not hold.
> 
> Please explain if I do not understand you correctly... (although maybe I should check out EAF terminal first)


Sorry I'm not good at English, look at this picture below, is this the effect you want?

https://ibb.co/TKtkLxR

loyalpartner:
> Yes thanks, I tried that too... but that did not work either... that was exactly what I was doing and what I pushed accidentally into this PR.

you add @interactive(insert_or_do=True) to copy_text,
you should bind key use this, i have test ok.
``` 
  (eaf-bind-key insert_or_copy_text "y" eaf-browser-keybinding)
```

dalanicolai:
> Sorry I'm not good at English, look at this picture below, is this the effect you want?
> 
> https://ibb.co/TKtkLxR

Yes this looks like, how I would like it...

loyalpartner:
i bind SPC use key-translation-map .
```
  (define-key key-translation-map (kbd "SPC")
    (lambda (prompt)
      (if (derived-mode-p 'eaf-mode)
          (pcase eaf--buffer-app-name
            ("browser" (if (eaf-call "call_function" eaf--buffer-id "is_focus")
                           (kbd "SPC")
                         (kbd "C-SPC")))
            ("pdf-viewer" (kbd "C-SPC"))
            ("image-viewer" (kbd "C-SPC"))
            (_  (kbd "SPC")))
        " ")))
```

dalanicolai:
> you add @Interactive(insert_or_do=True) to copy_text,
> you should bind key use this, i have test ok.
> 
> ```
>   (eaf-bind-key insert_or_copy_text "y" eaf-browser-keybinding)
> ```
As far as I know this is exactly what I was trying, but I got the message when pressing `y` that emacs could not run the function `insert_or_run_copy_text`. Well I am trying this is a Spacemacs layer, maybe some things get loaded at the wrong time. I will try it again then.

Anyway, great thanks for you help!

dalanicolai:
> i bind SPC use key-translation-map .

Ah, I did not know about that one yet. I will read about i :+1:

loyalpartner:
> > you add @Interactive(insert_or_do=True) to copy_text,
> > you should bind key use this, i have test ok.
> > ```
> >   (eaf-bind-key insert_or_copy_text "y" eaf-browser-keybinding)
> > ```
> 
> As far as I know this is exactly what I was trying, but I got the message when pressing `y` that emacs could not run the function `insert_or_run_copy_text`. Well I am trying this in a Spacemacs layer, maybe some things get loaded at the wrong time.
> So I will try it again then.
> 
> Anyway, great thanks for you help!

this is my step.
1. add @Interactive(insert_or_do=True) to copy_text
2.   (eaf-bind-key insert_or_copy_text "y" eaf-browser-keybinding) C-x C-e
3. M-x eaf-reastort-process

then try it.

PR #507, opened 2021-01-11
Do not ask for confirmation for apt and dnf commands in install-eaf

dalanicolai:
For the Spacemacs EAF layer I have created an `install-eaf-dependencies` command (see [here](https://github.com/syl20bnr/spacemacs/blob/077b8ea07cac2917a69db3cc84601573e54710a5/layers/%2Btools/eaf/funcs.el)). However, I have created the command such that it requires automatic confirmation. This should be safe for the apt and dnf package managers. For arch, I read that maybe the option --no-confirm, would do it safely, but I can not test that. So I have only added it for the apt and dnf commands.
I hope you agree to merge it, because it would be very nice for the layer.

### DIFF
--- a/install-eaf.sh
+++ b/install-eaf.sh
@@ -7,20 +7,20 @@ ARCH_PACKAGES+=(python-pyqt5 python-pyqt5-sip python-pyqtwebengine python-qrcode
 ARCH_PACKAGES+=(python-dbus python-pyinotify python-markdown python-qtconsole)
 
 # System dependencies
-if apt-get -v &> /dev/null; then
-    sudo apt-get install git nodejs aria2 libreoffice wmctrl xdotool
-    sudo apt-get install libglib2.0-dev libdbus-1-3 libdbus-1-dev
+if apt -v &> /dev/null; then
+    sudo apt -y install git nodejs aria2 libreoffice wmctrl xdotool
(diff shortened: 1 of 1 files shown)


MatthewZMD:
Thank you! Could you please create a PR for the `install-eaf-dependencies` function in the spacemacs layer? It'd be nice to integrate that function into EAF itself.

dalanicolai:
Sure!!

manateelazycat:
In arch, we need option --noconfirm to make pacman not ask confirmation, I have commit patch to fixed it.

BTW, can you help us confirm Fedora have package filebrowser-bin? Can you add it in install-eaf.sh if Fedora repository has this package?

Thanks.

dalanicolai:
> In arch, we need option --noconfirm to make pacman not ask confirmation, I have commit patch to fixed it.
> 
> BTW, can you help us confirm Fedora have package filebrowser-bin? Can you add it in install-eaf.sh if Fedora repository has this package?
> 
> Thanks.

Unfortunately I can not find this package in any Fedora repository

manateelazycat:
> Unfortunately I can not find the filebrowser-bin or any equivalent package in any Fedora repository

Okay, thank you for help.

PR #508, opened 2021-01-11
Add eaf-install-dependencies command and add quelpa recipe to doc

dalanicolai:
As installing via Quelpa works fine on Spacemacs, I have added an example Quelpa recipe to the documentation. I think, like in Spacemacs, also in vanilla Emacs Quelpa packages can be managed/updated via the Emacs package-manager.


MatthewZMD:
Awesome, many thanks!!

PR #516, opened 2021-01-14
Create insert_or_dark_mode command (i.e. add decorator)

dalanicolai:
This is necessary to activate dark_mode (without restarting eaf-process) from browser.

I could not find a way to deactivate dark-mode without restarting the eaf-process. Is there an easy way? (Otherwise I already have a function to do it, that requires restarting the eaf-process)

### DIFF
--- a/core/browser.py
+++ b/core/browser.py
@@ -577,7 +577,7 @@ Otherwise, scroll page up.
         ''' Clear the focus.'''
         self.eval_js(self.clear_focus_js)
 
-    @interactive()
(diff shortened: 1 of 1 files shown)


manateelazycat:
> This is necessary to activate dark_mode (without restarting eaf-process) from browser.
> 
> I could not find a way to deactivate dark-mode without restarting the eaf-process. Is there an easy way? (Otherwise I already have a function to do it, that requires restarting the eaf-process)

toggle dark mode and then call refresh_page.

PR #523, opened 2021-01-20
Add support eaf-search-it selection from pdf-view mode

dalanicolai:
Now eaf-search-it sets default search string to selection also in pdf-view mode


MatthewZMD:
Great!

PR #570, opened 2021-02-06
Readme typo: extraneous parentheses in Quelpa recipe

aspiwack:
The current Quelpa expression in the readme fails with a pretty confusing error, because of one too many pair of parentheses.


manateelazycat:
can you update README-zh_CN.md ？

aspiwack:
Here you go :slightly_smiling_face:

PR #574, opened 2021-02-08
normalize installation of node apps for different platform

vconcat:


MatthewZMD:
Great solution! Nice!

PR #578, opened 2021-02-10
Add option to control whether override Org store link function for PD…

stardiviner:
This will use docview: link type for EAF opened PDF by default.
Why use this link type by default? Because it's support in Org Mode built-in. And most compatible link type.
Might change this option into a string like "docview", "pdf-tools", "pdfview", etc. But later, I need to thinking.

Thanks for reviewing my PR.

### DIFF
--- a/eaf-org.el
+++ b/eaf-org.el
@@ -43,7 +43,7 @@
     (require 'org-docview)
   (require 'ol))
 
-(defcustom eaf-org-override-pdf-links nil
+(defcustom eaf-org-override-pdf-links-open nil
   "When enabled, this will override existing PDF file links's open function.
 
 So that every existing PDF org-link that's supposed to be opened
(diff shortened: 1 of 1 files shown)


manateelazycat:
@luhuaei Can you help to review this patch ?

luhuaei:
Thanks cc. Great work @stardiviner :smile:

stardiviner:
😄 I just improved my old code. Seems only me touched my own code. Only me use it...... lonely....

MatthewZMD:
>  I just improved my old code. Seems only me touched my own code. Only me use it...... lonely....

It's good code, I've demo-ed org-store-link at the EmacsConf2020 :)

stardiviner:
I'm happy that you use it. Happyyyy

PR #585, opened 2021-02-17
Improve markers.js on bing.com and weiyun.com

yqu212:
@MatthewZMD I have added several selectors to
make some links on cn.bing.com and weiyun.com clickable for eaf browser.
It's not perfect. Help wanted!

I notice that some codes in markders.js are from surfingkeys.
However, it is not as powerful as surfingkeys. The added selectors are
unnecessar for surfingkeys. I don't know why. I'm new to Javascript and
have not read through marker.js. Please have a look at this issue.

### DIFF
--- a/core/js/marker.js
+++ b/core/js/marker.js
@@ -47,7 +47,7 @@ try {
     }
 
     function isElementClickable(e) {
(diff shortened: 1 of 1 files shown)


MatthewZMD:
Thank you!

PR #586, opened 2021-02-17
prevent flash when loading new image in image-viewer

yqu212:


manateelazycat:
It's cool.

Can you fix flash when image-viewer load first picture too? ;)

yqu212:
Is there still a flash now? I'll have a look at it today.

PR #594, opened 2021-02-24
Don't use empty filename for newly mindmap

mackong:
Use " " as filename will create many random files caused by *auto save*, and also we can't rename the buffer. So specify the filename when creation should be better.


MatthewZMD:
That's right, thanks

PR #606, opened 2021-03-12
Make sure load finished before initialize

mackong:
Sometimes index.html not load finished, then js functions like `open_file`, `init_background` evaluated in `initialize` will failed, and the mindmap open failed.


manateelazycat:
Cool.

Issue #60, opened 2019-06-15
Not working with HiDPI


yangsheng6810:
I am using emacs 26.2 on i3 on Arch Linux. On one machine with normal display (1366x768 on a 14 inch screen), eaf works. However, on the other machine with HiDPI (2560x1440 on a 14 inch screen), eaf is kind of cropped, as shown in the following picture.
![1](https://user-images.githubusercontent.com/2863729/59547368-d73ca980-8f0b-11e9-9c8e-24cba9eaec7a.png)
To make things worse, the display from eaf totally covers the components from emacs, i.e. scroll bar, modeline, minibuffer, etc. It also blocks all inputs to emacs. I tried wrapping
```python
app = QApplication(sys.argv)
```
with 
```python
        os.environ["QT_AUTO_SCREEN_SCALE_FACTOR"] = "1"
        app = QApplication(sys.argv)
        app.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling)
```
as suggested by [pyat5 and 4k screen](https://stackoverflow.com/questions/41331201/pyqt-5-and-4k-screen), but it didn't work.

I managed to make it work with the following quick hack since I have a scale of 2. Ideally, this number should come from some variable or environmental variable. This can be tricky though: this might involve scaling both from the python side and the emacs side. Since currently PDF is showing a little bit of blur, I guess the proper way of handling would be emacs pass the exact available width/height, while PyQt handles the HiDPI case.
```diff
modified   eaf.el
@@ -221,6 +221,8 @@ We need calcuate render allocation to make sure no black border around render co
          (width (frame-pixel-width))
          ;; Render height should minus mode-line height, minibuffer height, header height.
          (height (- (frame-pixel-height) (window-mode-line-height) (window-pixel-height (minibuffer-window)) window-header-height)))
+    (setq width (/ width 2)
+          height (/ height 2))
     (mapcar (lambda (x) (format "%s" x)) (list width height))))
 
 (defun eaf-get-window-allocation (&optional window)
@@ -230,6 +232,10 @@ We need calcuate render allocation to make sure no black border around render co
          (w (- (nth 2 window-edges) x))
          (h (- (nth 3 window-edges) y))
          )
+    (setq x (/ x 2)
+          y (/ y 2)
+          w (/ w 2)
+          h (/ h 2))
     (list x y w h)))
 
 (defun eaf-generate-id ()
```
[...]

PR #613, opened 2021-03-19
Allow setting a scaling factor for HiDPI screens

leoc:
This takes the code from #60 and adds a scaling factor variable.

It tries to find the respective scaling factor by using the xrdb
settings from linux. For windows this should fall back to 1 and can be
set manually.

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -290,6 +290,14 @@ been initialized."
   "Name of EAF buffer."
   :type 'string)
 
+(defcustom eaf-scaling-factor
+  (/ (max 96
+	  (string-to-number
+	   (shell-command-to-string "xrdb -query | grep dpi | cut -d\":\" -f2 | tr -d \"[:space:]\"")))
+     96)
+  "Scaling factor of the Display for HiDPI displays."
+  :type 'integer)
+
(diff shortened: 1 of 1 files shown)


manateelazycat:
EAF can works with 4k screen，which os do you use?

leoc:
I am using Archlinux with AwesomeWM and the embedded x clients always were rendered two times the size of the frames width and height.

manateelazycat:
> I am using Archlinux with AwesomeWM and the embedded x clients always were rendered two times the size of the frames width and height.

EAF works well on gnome3/kde with 4k screen

can you test gnome3?

I believe something wrong in awesomewm config, not EAF

MatthewZMD:
> > I am using Archlinux with AwesomeWM and the embedded x clients always were rendered two times the size of the frames width and height.
> 
> EAF works well on gnome3/kde with 4k screen
> 
> can you test gnome3?
> 
> I believe something wrong in awesomewm config, not EAF

If this workaround works well for AwesomeWM, then we could probably accept it, given that it activates only in AwesomeWM

MatthewZMD:
Can you `git pull` to test the latest commit to see if it works for you?

manateelazycat:
I believe newest version has fix this issue, close PR

Issue #47, opened 2018-10-04
Lose emacs focus when using with EXWM


vuquangtam:
I use emacs25 in another xserver (run in Xephyr) with [exwn](https://github.com/ch11ng/exwm). When i try opening eaf, everytime i focus on eaf window, i lose my emacs focus (can't type on emacs).

manateelazycat:
EAF use complicated technology that depend on general desktop environment, such as, KDE/DDE/Unity/Gnome2

EAF won't work if you run it on any Tile WM or other light DE (LXDE, Xfce, OpenBox etc)

markgdawson:
Is it possible to have an idea of technology you're referring to? I'd be interested in trying to make this work if it was feasible.

markgdawson:
A quick test seems to fix the issue I was seeing. Would it be worth me sending a PR to add EXWM to the `eaf-wm-focus-fix-wms` list?

PR #618, opened 2021-03-25
Add EXWM to default eaf-wm-focus-fix-wms variable

markgdawson:
EXWM doesn't work out of the box, but adding to this variable seems to fix the issue (tested with the EAF browser).

This proposed change is discussed in #47.


manateelazycat:
can you add exwm info in README ?

PR #640, opened 2021-04-26
Fix eaf-install-dependencies on linux

### LINKED ISSUE #633: Installation via M-x eaf-install-dependencies does not work well
I experience two different problems with EAF on Ubuntu 20.10, one problem occurs in vanilla Emacs and the other problem occurs in Spacemacs. I am using EAF on Fedora 34 also, without any problem.

In this bug report I will describe the issue in Emacs. In a previous issue I have described the issue in Spacemacs.

**Describe the bug**
Okay I have found there is not a separate problem (from Spacemacs). I have just found out that installing the required packages using `eaf-install-dependencies` does not work correctly.

Manually running `install-eaf.sh` from the eaf directory installs things correctly.
Note that now I experience the same reparenting issue as I have just reported for Spacemacs in #632

Here follows the report as I was planning to write.

Subject: browser window stays blank
In Emacs when I open the browser with `M-x eaf-open-browser` then the browser does not find any webpage, but it stays blank.

It might be related to the following message that is shown in the \*epc\* buffer:
image

**To Reproduce**
Install eaf according to eaf installation instructions, and open the browser.

**Expected behavior**
Everything works fine

**Versions (please complete the following info):**
Ubuntu 20.10 with Gnome 3.38.3
 - Versions of Dependencies: newest versions available in Ubuntu default repo (let me know if you would like more specific info)
 - Emacs version 28.0.50

**Screenshots**
Irrelevant

**Additional context**
While creating this issue I realized that I am on Wayland, I will check the behavior on X11

**Additional context**

dalanicolai:
Don't use tramp with sudo because npm should run without sudo. The package
manager command is already sudoed within the install.sh script.

The PR additionally adds another required pip (PyQt5) dependency

This PR is expected to fix #633 (it is a little too much work to fully test it)


manateelazycat:
please do not install pyqt by pip, pip version has crash bug

dalanicolai:
Okay, I have moved PyQt5 it to get installed via `dnf install`

PR #645, opened 2021-04-29
Fix dnf install command

dalanicolai:
Sorry I have been very sloppy in #640. I have added the wrong package name. This commit fixes that...


MatthewZMD:
No worries, thanks!

PR #643, opened 2021-04-28
POSIX shell compatibility

lebensterben:
See #631

PR #649, opened 2021-04-30
fixup! #643

lebensterben:


lebensterben:
in #643 the line was `ARCH_PACKAGES=$(git nodejs npm)`
it's partially fixed in #644 `ARCH_PACKAGES=(git nodejs npm)`

But to be posix-compliant, array should not be used.

Issue #631, opened 2021-04-19
Multiple major issues with `install-eaf.sh`


lebensterben:
https://github.com/manateelazycat/emacs-application-framework/blob/b21efedaf78a2c41c36932c92884d930448b0516/install-eaf.sh#L1

- `/bin` is not standard PATH according to Linux Filesystem Specification.
- `bash` is not installed on some systems.
- One should always use `#!/usr/bin/env sh`

---

https://github.com/manateelazycat/emacs-application-framework/blob/b21efedaf78a2c41c36932c92884d930448b0516/install-eaf.sh#L5-L7

Array syntax is not available in POSIX shell. On system that `bash` or `zsh` is aliased to `sh`, this works. But on system with `dash`, this will not.

The solution is to store them in a string. And shell word split would separate them.

---

https://github.com/manateelazycat/emacs-application-framework/blob/b21efedaf78a2c41c36932c92884d930448b0516/install-eaf.sh#L10
https://github.com/manateelazycat/emacs-application-framework/blob/b21efedaf78a2c41c36932c92884d930448b0516/install-eaf.sh#L19
https://github.com/manateelazycat/emacs-application-framework/blob/b21efedaf78a2c41c36932c92884d930448b0516/install-eaf.sh#L28
https://github.com/manateelazycat/emacs-application-framework/blob/b21efedaf78a2c41c36932c92884d930448b0516/install-eaf.sh#L47
https://github.com/manateelazycat/emacs-application-framework/blob/b21efedaf78a2c41c36932c92884d930448b0516/install-eaf.sh#L49

- This redirection is only valid in bash. In posix shell you should use `COMMAND > /dev/null 2>&1`
- These tests are not canonical. You should use `[ $(command -v PROGRAM) ]`.

---

A more serious issue is, this script uses the existence of package manager to determine the host OS.
This is not robust.

On my system there is `dnf` but it can't install anything. It's a developer tool.

MatthewZMD:
Thanks for the report, the install script indeed needs some cleanup and improvement. It wasn't too long ago for it to be created, and the main goal was to "make it work" first.

Perhaps you could contribute by sending us a patch to improve this script!

manateelazycat:
@lebensterben Can you give me some PR to fix this issue directly ? Thanks.

PR #670, opened 2021-05-20
check node/npm installation status on ubuntu before install it.

chenzaichun:
Ubuntu has some deps issues if node is installed with node source
repoistory. check command is exists before apt install.

This is a workaround for npm deps issue if it had been installed manually.

see #666

### DIFF
--- a/install-eaf.sh
+++ b/install-eaf.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 set -eu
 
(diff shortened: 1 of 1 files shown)


install-eaf.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
manateelazycat, line 1: Can keep sh this place?

I remember some contributor change from bash to sh.

manateelazycat: @chenzaichun Check this.

chenzaichun: bash is installed on most of Linux distro and is the default shell. What shell should it support, ksh, csh, tcsh?

manateelazycat: Check https://github.com/manateelazycat/emacs-application-framework/issues/631

chenzaichun: Thanks. I'll update it later. by the way, As i know, $() is not support by all some shell, for example, csh is not support it.

chenzaichun: I'll remove array usage and command -v calling like #631.

manateelazycat: > I'll remove array usage and command -v calling like #631.

Nice ;)

PR #671, opened 2021-05-20
Fix/ubuntu npm issue

chenzaichun:
Rework with POSIX shell like #631


manateelazycat:
can you rebase and update?

thanks

PR #690, opened 2021-06-01
add eaf-open-pdf-with-history function

wcq062821:


eaf.el
```diff
@@ -2207,6 +2207,52 @@ If ALWAYS-NEW is non-nil, always open a new terminal for the dedicated DIR."
...
+                   (if history-file-exists
+                       (mapcar
+                        (lambda (h) (when (string-match history-pattern h)
+                                 (if (file-exists-p h)
+                                     (format "%s" h))))
+                        (with-temp-buffer (insert-file-contents pdf-history-file-path)
+                                          (split-string (buffer-string) "\n" t)))
+                     (make-directory (file-name-directory pdf-history-file-path) t)
+                     (f-write-text "" 'utf-8 pdf-history-file-path)))))
+    (if history-pdf (eaf-open history-pdf))))
```
MatthewZMD, line 2254: This should rename to `eaf-open-pdf-from-history`, it is not the same as `eaf-open-browser-with-history`, as the latter gives an option to search a keyword in a search engine, while this only enables opening PDF from history.

eaf.el
```diff
@@ -2221,6 +2267,7 @@ When called interactively, URL accepts a file that can be opened by EAF."
   (interactive "F[EAF] EAF Open: ")
   ;; Try to set app-name along with url when calling INTERACTIVELY
   (when (and (not app-name) (file-exists-p url))
+    (eaf-store-pdf-history url)
```
MatthewZMD, line 2270: Don't simply put it here, wrap it with a condition to check if it's opening a PDF, and create a variable for user to enable/disable this feature.

wcq062821: 好的，  已经修复。由于存储PDF文件路径并不会占用多大存储空间 所以我默认是开启的， 不知道这样可以吗？

MatthewZMD: > 所以我默认是开启的， 不知道这样可以吗？

主要Emacs用户群体有很多人对隐私比较敏感，加一下比较好

PR #715, opened 2021-07-21
eaf-pdf-viewer: Imenu support.

c1-g:
To test, call `M-x imenu` on a buffer opened with EAF’s pdf-viewer.
If the file has a table of contents, it will have you select a chapter
or whatever your file’s outline is for example, “Chapter 1”,
“Preface” or “Glossary”. Otherwise, it will have you select
a page from all the pages your file has in a generic “Page 1”, “Page
2”, etc.

Tips: the selection candidates look a little wacky since it’s the default’s
imenu. Try calling `helm-imenu` or `counsel-imenu` instead, the results
will look much nicer.


MatthewZMD:
Thanks for the patch!

PR #721, opened 2021-07-22
eaf-pdf-viewer: New eaf-pdf-outline-view

c1-g:
The new `eaf-pdf-outline-view` is like `eaf-pdf-outline-jump` but without switching to the pdf window.

### DIFF
--- a/app/pdf-viewer/eaf-pdf-viewer.el
+++ b/app/pdf-viewer/eaf-pdf-viewer.el
@@ -104,12 +104,14 @@
   :keymap (let ((map (make-sparse-keymap)))
             (define-key map (kbd "RET") 'eaf-pdf-outline-jump)
             (define-key map (kbd "q") 'quit-window)
+            (define-key map (kbd "o") 'eaf-pdf-outline-view)
+            (define-key map (kbd "v") 'eaf-pdf-outline-view)
             map))
 
 (defun eaf-pdf-outline ()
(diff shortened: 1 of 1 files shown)


app/pdf-viewer/eaf-pdf-viewer.el
```diff
@@ -119,12 +121,14 @@
...
       (erase-buffer)
+      (setq outline-regexp "^[[:space:]]*")
       (insert toc)
+      (outline-mode)
       (setq toc (mapcar (lambda (line)
                           (string-to-number (car (last (split-string line " ")))))
                         (butlast (split-string (buffer-string) "\n"))))
       (goto-line (seq-count (apply-partially #'>= page-number) toc))
-      (set (make-local-variable 'eaf-pdf-outline-original-buffer-name) buffer-name)
+      (set (make-local-variable 'eaf-pdf-outline-original-buffer-name) buf)
```
manateelazycat, line 131: this change is correct?

c1-g: Yes, I changed `eaf-pdf-outline-original-buffer-name` to be a buffer object rather than its name. Should've changed the variable name to reflect that. This is important since `eaf-pdf-outline-view` use the buffer object to find `eaf--buffer-id` of the original buffer

MatthewZMD:
Thanks for the patch!

PR #737, opened 2021-07-29
pdf-viewer: A more useful mode line position indicator.

c1-g:
The old position indicator always displays `All (1,0)` which is not very useful. This will replace that to display the current page and the total page of the document.

There are two local variables added; `eaf-pdf-current-page` and `eaf-pdf-total-page`,
the current-page variable is set with `set_emacs_var` on the Python's side every time user scrolls.

### DIFF
--- a/app/pdf-viewer/buffer.py
+++ b/app/pdf-viewer/buffer.py
@@ -26,7 +26,7 @@ from PyQt5.QtGui import QPainter
 from PyQt5.QtWidgets import QWidget
 from PyQt5.QtWidgets import QToolTip
 from core.buffer import Buffer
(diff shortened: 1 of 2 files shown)


manateelazycat:
This patch is too complicated, please check newest commit https://github.com/manateelazycat/emacs-application-framework/commit/ac1664a4e56eaedb62aa717d94ef03e15c5ea834

If you just want call some function when pdf vertical offset change, you just need write implement in function ```eaf--pdf-update-position```.

You don't need set emacs variable at Python side, add two customize options, and run hooks? Why?

Why not call some elisp function with page-index and page-total-number when offset change?

PR #739, opened 2021-07-30
pdf-viewer: Mode line position.

c1-g:
A simplified version of #737 based on commit ac1664a.


MatthewZMD:
Please test your code before creating a PR - there is an extra ) at the end of the function that breaks the entire EAF.

PR #744, opened 2021-07-31
Declare eaf-setq as obsolete function.

c1-g:
Users that use `eaf-setq` in their configuration will see a `void-function` error so this will fix it.


MatthewZMD:
I also updated in the [discussion](https://github.com/manateelazycat/emacs-application-framework/discussions/527), thanks!

PR #745, opened 2021-08-01
Autoload all open-* function

c1-g:
`require`ing all these plugins make loading eaf.el slow.  And moving `eaf-open-*` functions to their own file makes navigation easier.

Some applications that don't have their own open function will still be loaded as library. (`pdf-viewer`, `js-video-player`, `video-player`, `markdown-previewer`)

I also created .el files for `airshare`, `file-sender`.


manateelazycat:
Awesome patch, thanks!

PR #747, opened 2021-08-01
[WIP] Autoload eaf-open-* function

c1-g:
I made a mistake in #745 to autoload files without adding them to load-path first, somehow I couldn't revert the commit so I have to create a new one. 

Work-in-progress because I still need others to test this, for me it's working fine.


MatthewZMD:
Rebase with master, it has been fixed.

manateelazycat:
Newest version has fixed this issue.

We can't use autload syntax now, because some app function still use in eaf.el

PR #773, opened 2021-08-13
Use pygetwindow instead of vbs to activate win32 window

xhcoding:


manateelazycat:
@xhcoding I haven't found pygetwindow, miss commit ? ;)

xhcoding:
in eaf.py line 43   ;)

```python
if platform.system() == "Windows":
    import pygetwindow as gw
```

manateelazycat:
I see, thanks!

PR #795, opened 2021-08-17
[draft] Do our best to avoid background change like:  dark->light-> dark or light->dark->light.

tumashu:


manateelazycat:
Thsi patch is not right.

PR #796, opened 2021-08-17
display hovered link and Ctrl+MouseWheel to open link in new window

taquangtrung:
Hi,

I implemented in this PR the feature to:
- display hovered links in Emacs' status bar, and 
- press Ctrl + click MouseWheel button to open links in new window.

Can you check if this is useful?

Thanks!

### DIFF
--- a/core/webengine.py
+++ b/core/webengine.py
@@ -35,14 +35,14 @@ import os
 import platform
 import sqlite3
 
+MOUSE_LEFT_BUTTON = 1
+MOUSE_WHEEL_BUTTON = 4
 MOUSE_BACK_BUTTON = 8
 MOUSE_FORWARD_BUTTON = 16
-MOUSE_WHEEL_BUTTON = 4
 
 class BrowserView(QWebEngineView):
 
     translate_selected_text = QtCore.pyqtSignal(str)
-    url_hovered = ""
 
     def __init__(self, buffer_id):
         super(QWebEngineView, self).__init__()
@@ -58,7 +58,8 @@ class BrowserView(QWebEngineView):
         self.cookie_storage = BrowserCookieStorage(self.config_dir)
         self.cookie_store.cookieAdded.connect(self.cookie_storage.add_cookie)
 
-        self.page().linkHovered.connect(self.linkHovered)
+        self.url_hovered = ""
+        self.page().linkHovered.connect(self.link_hovered)
 
         self.selectionChanged.connect(self.select_text_change)
 
@@ -253,11 +254,16 @@ class BrowserView(QWebEngineView):
                 event.accept()
                 return True
 
-            elif event.button() == MOUSE_WHEEL_BUTTON:
+            elif event.button() == MOUSE_LEFT_BUTTON:
                 modifiers = QApplication.keyboardModifiers()
                 if modifiers == Qt.ControlModifier:
                     self.open_url_new_buffer_other_window(self.url_hovered)
                     return True
+
+            elif event.button() == MOUSE_WHEEL_BUTTON:
+                self.open_url_new_buffer_other_window(self.url_hovered)
+                return True
+
 
         if event.type() == QEvent.Wheel:
             modifiers = QApplication.keyboardModifiers()
@@ -269,7 +275,7 @@ class BrowserView(QWebEngineView):
 
         return super(QWebEngineView, self).eventFilter(obj, event)
 
-    def linkHovered(self, url):
+    def link_hovered(self, url):
         if url:
             self.url_hovered = url
             message_to_emacs(url)
--- a/core/webengine.py
+++ b/core/webengine.py
@@ -37,10 +37,12 @@ import sqlite3
 
 MOUSE_BACK_BUTTON = 8
 MOUSE_FORWARD_BUTTON = 16
+MOUSE_WHEEL_BUTTON = 4
 
 class BrowserView(QWebEngineView):
 
     translate_selected_text = QtCore.pyqtSignal(str)
+    url_hovered = ""
 
     def __init__(self, buffer_id):
         super(QWebEngineView, self).__init__()
@@ -55,6 +57,8 @@ class BrowserView(QWebEngineView):
         self.cookie_store = self.page().profile().cookieStore()
         self.cookie_storage = BrowserCookieStorage(self.config_dir)
         self.cookie_store.cookieAdded.connect(self.cookie_storage.add_cookie)
+
+        self.page().linkHovered.connect(self.linkHovered)
 
         self.selectionChanged.connect(self.select_text_change)
 
@@ -249,6 +253,12 @@ class BrowserView(QWebEngineView):
                 event.accept()
                 return True
 
+            elif event.button() == MOUSE_WHEEL_BUTTON:
+                modifiers = QApplication.keyboardModifiers()
+                if modifiers == Qt.ControlModifier:
+                    self.open_url_new_buffer_other_window(self.url_hovered)
+                    return True
+
         if event.type() == QEvent.Wheel:
             modifiers = QApplication.keyboardModifiers()
             if modifiers == Qt.ControlModifier:
@@ -258,6 +268,12 @@ class BrowserView(QWebEngineView):
                     self.zoom_out()
 
         return super(QWebEngineView, self).eventFilter(obj, event)
+
+    def linkHovered(self, url):
+        if url:
+            self.url_hovered = url
+            message_to_emacs(url)
+        return True
 
     def open_url(self, url):
         ''' Configure current url.'''
--- a/core/webengine.py
+++ b/core/webengine.py
@@ -27,7 +27,7 @@ from PyQt5.QtWebEngineWidgets import QWebEngineView, QWebEnginePage, QWebEngineS
 from PyQt5.QtWidgets import QApplication, QWidget
 from PyQt5.QtWebChannel import QWebChannel
 from core.buffer import Buffer
-from core.utils import touch, string_to_base64, popen_and_call, call_and_check_code, interactive, abstract, eval_in_emacs, message_to_emacs, open_url_in_background_tab, duplicate_page_in_new_tab, open_url_in_new_tab, open_url_in_new_tab_other_window, focus_emacs_buffer, atomic_edit, get_emacs_var, get_emacs_config_dir, to_camel_case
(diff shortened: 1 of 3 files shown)


core/webengine.py
```diff
@@ -37,10 +37,12 @@
 
 MOUSE_BACK_BUTTON = 8
 MOUSE_FORWARD_BUTTON = 16
+MOUSE_WHEEL_BUTTON = 4
 
 class BrowserView(QWebEngineView):
 
     translate_selected_text = QtCore.pyqtSignal(str)
+    url_hovered = ""
```
manateelazycat, line 45: No, this is area to write signal, please don't define class variable, you should define instance variable.

Looks next commit.

core/webengine.py
```diff
@@ -56,6 +58,8 @@ def __init__(self, buffer_id):
         self.cookie_storage = BrowserCookieStorage(self.config_dir)
         self.cookie_store.cookieAdded.connect(self.cookie_storage.add_cookie)
```
manateelazycat, line 60: You can define ```self.hovered_url``` before linkHovered signal connect.

core/webengine.py
```diff
@@ -259,6 +269,12 @@ def eventFilter(self, obj, event):
 
         return super(QWebEngineView, self).eventFilter(obj, event)
 
+    def linkHovered(self, url):
```
manateelazycat, line 272: Python function should use underline style, like ```link_hovered```, camel style only use if library use camel style, such as PyQt

core/webengine.py
```diff
@@ -249,6 +253,12 @@ def eventFilter(self, obj, event):
                 event.accept()
                 return True
 
+            elif event.button() == MOUSE_WHEEL_BUTTON:
+                modifiers = QApplication.keyboardModifiers()
+                if modifiers == Qt.ControlModifier:
+                    self.open_url_new_buffer_other_window(self.url_hovered)
+                    return True
+
         if event.type() == QEvent.Wheel:
```
manateelazycat, line 262: Why not use Ctrl + Button?
Ctrl + button is more general for user.

manateelazycat (changes requested):

manateelazycat:
@taquangtrung Thanks for patch, please check my comments.

core/webengine.py
taquangtrung: Currently, Ctrl+Button and MouseWheel will open links in new buffers but in the same window as the current page.
Do you think it's ok to override this feature?

taquangtrung: On second thought, I agree with you and change the code to make both Ctrl+LeftButton and WheelButton to open links in new window.

core/webengine.py
taquangtrung: thanks!

taquangtrung:
@manateelazycat  I just pushed the new patch.

However, I notice a problem that after open links in new windows, the mouse pointer suddenly jumps to the bottom-left corner of the monitor. 

This problem also occurs when I close a webpage by pressing `x`, so this isn't something related to my code.

Do you know what happens with the mouse poitner?

manateelazycat:
I guess you use some Tile Window Manager, please check document of eaf--activate-emacs-linux-window

taquangtrung:
Thanks! I will check that function. Currently, I'm using the standard window manager of Linux Mint 20.2 XFCE.

taquangtrung:
> I guess you use some Tile Window Manager, please check document of eaf--activate-emacs-linux-window and eaf-wm-focus-fix-wms

Great! I removed `Xfwm4` from `eaf-wm-focus-fix-wms` and the mouse pointer works nicely. 

Not sure why it is added to this list? :D

manateelazycat:
> > I guess you use some Tile Window Manager, please check document of eaf--activate-emacs-linux-window and eaf-wm-focus-fix-wms
> 
> Great! I removed `Xfwm4` from `eaf-wm-focus-fix-wms` and the mouse pointer works nicely.
> 
> Not sure why it is added to this list? :D

Because when you do Alt + Tab between Emacs/EAF and other application, EAF maybe not accept input event.

taquangtrung:
> > Great! I removed `Xfwm4` from `eaf-wm-focus-fix-wms` and the mouse pointer works nicely.
> > Not sure why it is added to this list? :D
> 
> Because when you do Alt + Tab between Emacs/EAF and other application, EAF maybe not accept input event.

Oh, thanks! I will check if that happens with my Emacs.

core/webengine.py
```diff
@@ -249,6 +254,17 @@ def eventFilter(self, obj, event):
                 event.accept()
                 return True
 
+            elif event.button() == MOUSE_LEFT_BUTTON:
+                modifiers = QApplication.keyboardModifiers()
+                if modifiers == Qt.ControlModifier:
+                    self.open_url_new_buffer_other_window(self.url_hovered)
+                    return True
+
+            elif event.button() == MOUSE_WHEEL_BUTTON:
```
manateelazycat, line 263: Please add Ctrl modifier check, if only check wheel event, user will trigger open new window when he scroll page after hover something.

manateelazycat (changes requested):

core/webengine.py
taquangtrung: Hmmm, I don't encounter this problem. I can scroll normally when hovering over the link. Only when the WheelButton is pressed, then the link will be opened (this is similar to Google Chrome's feature).

Can you double-check if this problem occurs in your machine?

taquangtrung:
@manateelazycat I just updated the code to clear Emacs' echo area when the mouse pointer doesn't hover over links.
I also add a condition to ensure that a link is opened only when it is being hovered at the exact moment the mouse button is clicked.

manateelazycat:
thanks for patch, awesome

taquangtrung:
> > > I guess you use some Tile Window Manager, please check document of eaf--activate-emacs-linux-window and eaf-wm-focus-fix-wms
> > 
> > 
> > Great! I removed `Xfwm4` from `eaf-wm-focus-fix-wms` and the mouse pointer works nicely.
> > Not sure why it is added to this list? :D
> 
> Because when you do Alt + Tab between Emacs/EAF and other application, EAF maybe not accept input event.

@manateelazycat: I just figured out that for Xfwm4 desktop, EAF doesn't need to move the mouse pointer to the bottom right of the screen to gain focus for Emacs.

Instead, I can just make EAF call a function in `buffer.py`, which basically does nothing (it just returns True), and EAF apps can still gain focus properly (although I have no idea why?).

Here is the idea of my patch:

- In the module `eaf/core/buffer.py`

  ```python

    def do_nothing(self):
        '''
        Do nothing
        '''
        return True
  ```

- In the module: `eaf.el`:

  ```emacs-lisp
  (defun eaf--activate-emacs-linux-window (&optional buffer_id)
    "Activate Emacs window by `wmctrl'."
    (if (member (eaf--get-current-desktop-name) eaf-wm-focus-fix-wms)
        ;; When switch app focus in WM, such as, i3 or qtile.
        ;; Emacs window cannot get the focus normally if mouse in EAF buffer area.
        ;;
        ;; So we move mouse to frame bottom of Emacs, to make EAF receive input event.

        ;; <<<<<<<<<<<<<<<<<<<<<<<
        ;; TRUNG: here is may patch to call "do_nothing", instead of "move_cursor_to_corner"
        (eaf-call-async "execute_function" (or eaf--buffer-id buffer_id) "do_nothing" (key-description (this-command-keys-vector)))
        ;; >>>>>>>>>>>>>>>>>>>>>>>

      ;; When press Alt + Tab in DE, such as KDE.
      ;; Emacs window cannot get the focus normally if mouse in EAF buffer area.
      ;;
      ;; So we use wmctrl activate on Emacs window after Alt + Tab operation.
      (if (executable-find "wmctrl")
          (shell-command-to-string (format "wmctrl -i -a $(wmctrl -lp | awk -vpid=$PID '$3==%s {print $1; exit}')" (emacs-pid)))
        (message "Please install wmctrl to active Emacs window."))))
  ```

I tested it in my Linux Mint 20.2 Xfce and it works perfectly fine. 

Would you like me to make a PR so that you can test it on your side?

manateelazycat:
> > > > I guess you use some Tile Window Manager, please check document of eaf--activate-emacs-linux-window and eaf-wm-focus-fix-wms
> > > 
> > > 
> > > Great! I removed `Xfwm4` from `eaf-wm-focus-fix-wms` and the mouse pointer works nicely.
> > > Not sure why it is added to this list? :D
> > 
> > 
> > Because when you do Alt + Tab between Emacs/EAF and other application, EAF maybe not accept input event.
> 
> @manateelazycat: I just figured out that for Xfwm4 desktop, EAF doesn't need to move the mouse pointer to the bottom right of the screen to gain focus for Emacs.
> 
> Instead, I can just make EAF call a function in `buffer.py`, which basically does nothing (it just returns True), and EAF apps can still gain focus properly (although I have no idea why?).
> 
> Here is the idea of my patch:
> 
> * In the module `eaf/core/buffer.py`:
>   ```python
>     def do_nothing(self):
>         '''
>         Do nothing
>         '''
>         return True
>   ```
> * In the module: `eaf.el`:
>   ```emacs-lisp
>   (defun eaf--activate-emacs-linux-window (&optional buffer_id)
>     "Activate Emacs window by `wmctrl'."
>     (if (member (eaf--get-current-desktop-name) eaf-wm-focus-fix-wms)
>         ;; When switch app focus in WM, such as, i3 or qtile.
>         ;; Emacs window cannot get the focus normally if mouse in EAF buffer area.
>         ;;
>         ;; So we move mouse to frame bottom of Emacs, to make EAF receive input event.
>   
>         ;; <<<<<<<<<<<<<<<<<<<<<<<
>         ;; TRUNG: here is my patch to call "do_nothing", instead of "move_cursor_to_corner"
>         (eaf-call-async "execute_function" (or eaf--buffer-id buffer_id) "do_nothing" (key-description (this-command-keys-vector)))
>         ;; >>>>>>>>>>>>>>>>>>>>>>>
>   
>       ;; When press Alt + Tab in DE, such as KDE.
>       ;; Emacs window cannot get the focus normally if mouse in EAF buffer area.
>       ;;
>       ;; So we use wmctrl activate on Emacs window after Alt + Tab operation.
>       (if (executable-find "wmctrl")
>           (shell-command-to-string (format "wmctrl -i -a $(wmctrl -lp | awk -vpid=$PID '$3==%s {print $1; exit}')" (emacs-pid)))
>         (message "Please install wmctrl to active Emacs window."))))
>   ```
> 
> I tested it in my Linux Mint 20.2 Xfce and it works perfectly fine.
> 
> Would you like me to make a PR so that you can test it on your side?

Sure, move_cursor is hacking way, anyway if do_nothing can get Qt window focus, it's a better patch!

PR #802, opened 2021-08-18
Ctrl + Mouse Back/Forward to switch to next/previous webpage buffer

taquangtrung:
Hi,

I implemented this feature to allow using Ctrl + Mouse Backward/Forward buttons to switch to the previous/next webpage buffers in the same Emacs window. 

It is different from the current Mouse Backward/Forward feature in that the latter only switch pages of the same buffer.

It is useful when a user wants to open many webpage buffers and navigate between them.

Could you advise if it is necessary to be included in EAF?

Thanks!


manateelazycat:
@taquangtrung Good patch, thanks!

PR #805, opened 2021-08-19
find focusing buffer on all visible frames

taquangtrung:
Hi,

I created this patch to fix a bug that EAF gets focus on the wrong window buffer when a user who is
- running Emacs in daemon mode with 2 frames
- working in one active frame
- and wants to click on a webpage link in the other frame.

The current code only finds the window buffer `(get-buffer-window buffer)` on the selected frame (the first frame). This isn't the correct frame, hence EAF cannot get the correct focus on the other frame containing the webpage.

I updated the new code to `(get-buffer-window buffer 'visible)` to enable EAF to find in the other frame as well.

Could you advise if this is a suitable fix?

Thanks!


manateelazycat:
cool

PR #811, opened 2021-08-19
gain EAF buffer focus by do_nothing, instead of move_cursor_to_corner

taquangtrung:
Hi,

I created this PR based on our discussion here: https://github.com/emacs-eaf/emacs-application-framework/pull/796#issuecomment-901762463

### DIFF
--- a/core/buffer.py
+++ b/core/buffer.py
@@ -179,6 +179,12 @@ class Buffer(QGraphicsScene):
         screen = qApp.primaryScreen()
         QCursor().setPos(screen, screen.size().width(), screen.size().height())
 
+    def do_nothing(self):
+        '''
+        Do nothing
+        '''
+        return True
+
     def set_aspect_ratio(self, aspect_ratio):
         ''' Set aspect ratio.'''
         self.aspect_ratio = aspect_ratio
--- a/eaf.el
+++ b/eaf.el
@@ -1372,20 +1372,11 @@ So multiple EAF buffers visiting the same file do not sync with each other."
 
 (defun eaf--activate-emacs-linux-window (&optional buffer_id)
   "Activate Emacs window by `wmctrl'."
-  (if (member (eaf--get-current-desktop-name) eaf-wm-focus-fix-wms)
-      ;; When switch app focus in WM, such as, i3 or qtile.
-      ;; Emacs window cannot get the focus normally if mouse in EAF buffer area.
-      ;;
-      ;; So we move mouse to frame bottom of Emacs, to make EAF receive input event.
-      (eaf-call-async "execute_function" (or eaf--buffer-id buffer_id) "do_nothing" (key-description (this-command-keys-vector)))
-
-    ;; When press Alt + Tab in DE, such as KDE.
-    ;; Emacs window cannot get the focus normally if mouse in EAF buffer area.
-    ;;
-    ;; So we use wmctrl activate on Emacs window after Alt + Tab operation.
-    (if (executable-find "wmctrl")
-        (shell-command-to-string (format "wmctrl -i -a $(wmctrl -lp | awk -vpid=$PID '$3==%s {print $1; exit}')" (emacs-pid)))
-      (message "Please install wmctrl to active Emacs window."))))
+  ;; try to call to `do_nothing` from an EAF buffer first to gain window focus
+  (eaf-call-async "execute_function" (or eaf--buffer-id buffer_id) "do_nothing" (key-description (this-command-keys-vector)))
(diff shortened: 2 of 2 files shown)


manateelazycat:
Maybe we should impolement ```eaf--activate-emacs-linux-window``` like below then delete ```eaf-wm-focus-fix-wms``` together?

```elisp
(defun eaf--activate-emacs-linux-window (&optional buffer_id)
  "Activate Emacs window by `wmctrl'."
  (eaf-call-async "execute_function" (or eaf--buffer-id buffer_id) "do_nothing" (key-description (this-command-keys-vector)))
  (when (executable-find "wmctrl")
    (shell-command-to-string (format "wmctrl -i -a $(wmctrl -lp | awk -vpid=$PID '$3==%s {print $1; exit}')" (emacs-pid)))))
```

If user report EAF can't accept focus after this patch, then we rever this patch.

taquangtrung:
> Maybe we should implemenet `eaf--activate-emacs-linux-window` like below then delete `eaf-wm-focus-fix-wms` together?
> 
> ```emacs-lisp
> (defun eaf--activate-emacs-linux-window (&optional buffer_id)
>   "Activate Emacs window by `wmctrl'."
>   (eaf-call-async "execute_function" (or eaf--buffer-id buffer_id) "do_nothing" (key-description (this-command-keys-vector)))
>   (when (executable-find "wmctrl")
>     (shell-command-to-string (format "wmctrl -i -a $(wmctrl -lp | awk -vpid=$PID '$3==%s {print $1; exit}')" (emacs-pid)))))
> ```
> 
> If user report EAF can't accept focus after this patch, then we rever this patch.

Sure, I just created a new commit

manateelazycat:
great patch!

c1-g:
Okay, so I've tested this on EXWM and it won't return focus to Emacs because somehow, `wmctrl -lp`  just doesn't list the emacs process.

taquangtrung:
> Okay, so I've tested this on EXWM and it won't return focus to Emacs because somehow, `wmctrl -lp` just doesn't list the emacs process.

Hmmm, I don't use EXWM hence I can't really test it. 

If possible, could you take a screenshot of your Emacs on EXWM and the output of `wmctrl -lp`?

manateelazycat:
> > Okay, so I've tested this on EXWM and it won't return focus to Emacs because somehow, `wmctrl -lp` just doesn't list the emacs process.
> 
> @c1-g : Hmmm, I don't use EXWM hence I can't really test it.
> 
> If possible, could you take a screenshot of your Emacs on EXWM and the output of `wmctrl -lp`?

EXWM use Emacs as window manager.

If we open two Emacs frame, one for EXWM, and another for EAF, focus is okay.

I think EXWM is conflict focus with EAF.

manateelazycat:
Xfce and LXDE is okay with newest version, i3 and QTile haven't test.

c1-g:
EXWM

@taquangtrung So here's the screen capture of EXWM managing one process of qutebrowser

On the `*Proced*` buffer you will see the list of all processes that belongs to user. and on *eshell* is what `wmctrl -lp` actually listed.

manateelazycat:
After developer test.

This patch is not work, because Xfce and LXDE works well and don't need move cursor to corner.

```
KDE, Gnome2, Gnome3, Mate, Xfce, LXDE, Sway works well default,
i3, QTile, Xpra, EXWM need move cursor to corner to get focus of Window manager.
```

I have remove Xfce window manager from eaf-wm-focus-fix-wms in newest version.

taquangtrung:
> After developer test.
> 
> This patch is not work, because Xfce and LXDE works well and don't need move cursor to corner.
> 
> ```
> KDE, Gnome2, Gnome3, Mate, Xfce, LXDE, Sway works well default,
> i3, QTile, Xpra, EXWM need move cursor to corner to get focus of Window manager.
> ```
> 
> I have remove Xfce window manager from eaf-wm-focus-fix-wms in newest version.

Hmmm, the old code doesn't work properly for me in XFCE when I open 2 Emacs frames (daemon mode).

If I'm using EAF in the secondary frame (not the primary frame), then by calling `wmctrl`, it will only activate the primary frame.

I still need to call to `do_nothing` as a way to gain focus on the secondary frame.

manateelazycat:
do_nothing is wrong, you perhaps need figure out wmctrl works with two frame.

taquangtrung:
> do_nothing is wrong, you perhaps need figure out wmctrl works with two frame.

I see, I will figure it out if I can pass window ID to `wmctrl` as well

PR #823, opened 2021-08-22
Add dependecies for installing on Opensuse with zypper

MKhashaki:


MatthewZMD:
Thanks for the patch!

PR #827, opened 2021-08-24
allow edit_url to search string

taquangtrung:
Hi,

I created this PR which allows users to search for a string when editing URL (`e`).

Currently, when a user presses `e` to edit the URL and just type `foo`, then EAF browser will return a blank page.
Using this PR, EAF can allow use to search for the string `foo` by using the default search engine.

This PR is accompanied by the PR https://github.com/emacs-eaf/eaf-browser/pull/9 in EAF browser.

Here, I need to create a new function `eval_get_result_in_emacs` to get the search string created by EAF browser.

Could you advise if this feature is useful and can be merged?

Thanks!

PR #830, opened 2021-08-25
support evaluating and getting result from an Emacs function

taquangtrung:
Hi,

I created this PR related to our discussion in the old PR: https://github.com/emacs-eaf/emacs-application-framework/pull/827

Here, I still need to add the function `eval_get_result_in_emacs` to `utils.py` so that `eaf-browser/buffer.py` can get the search string generated by `eaf-browser.el`.

### DIFF
--- a/core/utils.py
+++ b/core/utils.py
@@ -233,6 +233,21 @@ def eval_in_emacs(method_name, args):
         # Call eval-in-emacs elisp function.
         epc_client.call("eval-in-emacs", args)
 
+def eval_get_result_in_emacs(method_name, args):
+    global epc_client
+
+    if epc_client == None:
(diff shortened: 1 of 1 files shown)


core/utils.py
```diff
@@ -233,6 +233,21 @@ def eval_in_emacs(method_name, args):
         # Call eval-in-emacs elisp function.
         epc_client.call("eval-in-emacs", args)
 
+def eval_get_result_in_emacs(method_name, args):
```
manateelazycat, line 236: Maybe ```get_emacs_func_result``` is better name?

manateelazycat (changes requested):

core/utils.py
taquangtrung: Thanks! I just updated it.

PR #860, opened 2021-10-12
clear both searched text and selected text at the same time by quit-action

taquangtrung:
Hi,

This is a small tweak to `C-g` to cleared both searched text and selected text at the same time.

The current behavior of webengine is that when a user searches a text using
`C-s` and then presses `C-g` to quit, then the current searched text will still
highlighted (by the selection mode), while other nearby searched texts are cleared.

So, the user needs to press `C-g` again to clear that current searched text.

I put a timeout of 0.1s to wait for the current searched text to be highlighted,
then the next code will clear it:

``` python
        try:
            self.triggerPageAction(self.web_page.Unselect)
        except:
            pass
```

In the above code, I remove the condition `if self.web_page.hasSelection()`
since it is wrapped by the exception checking anyway, so that condition isn't
needed.

Can you review and consider if it can be merged?

Thanks!

### DIFF
--- a/core/webengine.py
+++ b/core/webengine.py
@@ -32,6 +32,7 @@ from functools import partial
 from urllib.parse import urlparse, parse_qs, urlunparse, urlencode
 import base64
 import os
+import time
 import platform
 import sqlite3
 
@@ -190,6 +191,8 @@ class BrowserView(QWebEngineView):
         ''' Quit action.'''
         if self.search_term != "":
             self._search_text("")
+            time.sleep(0.1)
+
         if self.buffer.caret_browsing_mode:
             if self.buffer.caret_browsing_mark_activated:
                 self.buffer.caret_toggle_mark()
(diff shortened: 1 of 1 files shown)


taquangtrung:
Ok, here is an example where I search for the text "virtual"

search1

Now, I press `C-g` for the first time (no `sleep` is invoked), then `self._search_text("")` will trigger a selection changed event, and it leads to the text selection as in the below screenshot:

search2

So, I have to press `C-g` again (2nd time) to clear the selected text.

I don't know why `self._search_text("")` will trigger such a selection event. 

So I put the sleep command to wait for the selection event to complete, and the code `self.triggerPageAction(self.web_page.Unselect)` can clear It too.

If we don't put the `sleep` command, then `self.triggerPageAction(self.web_page.Unselect)` is invoked to quickly, even before the time that the selection event completes.

manateelazycat:
> @manateelazycat: Oh no, I think the code `self.triggerPageAction(self.web_page.Unselect)` is still needed in `action_quit` to clear the selected text when you manually use the mouse the select it.
> 
> Now, we have two cases that can introduce selected text: when the user searches for a text (`C-g`) or uses the mouse to select a text.
> 
> Your new patch just handles the first case, but not the second case.

I just commit new patch https://github.com/emacs-eaf/emacs-application-framework/commit/27b45f47dac6772f4ec8551ea748c092e941d694

It good for you ?

taquangtrung:
@manateelazycat : I just created the PR for your convenience when reviewing the code: https://github.com/emacs-eaf/emacs-application-framework/pull/861

PR #862, opened 2021-10-13
refactor code using eaf-get-buffer

taquangtrung:
Hi,

I refactor the code to use the function `eaf-get-buffer` following the discussion here: https://github.com/emacs-eaf/emacs-application-framework/pull/859#issuecomment-941825924

Can you take a look to check if it is OK?

Thanks!


MatthewZMD:
Awesome!

PR #863, opened 2021-10-19
allow editing searched text to continue with the search (using M-e)

taquangtrung:
Hi, 

I created this PR to allow Eaf-browser to edit the search text (similar to pressing `M-e` in `isearch`).

Can you help to review and merge it?

Thanks!

### DIFF
--- a/core/webengine.py
+++ b/core/webengine.py
@@ -68,6 +68,8 @@ class BrowserView(QWebEngineView):
         self.load_cookie()
 
         self.search_term = ""
+        self.search_mode_forward = False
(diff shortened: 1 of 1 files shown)


manateelazycat:
Thanks for patch.

I think right way is need re-implement search feature like eaf-file-manager, then we can search anything in realtime and edit search text in minibuffer.

taquangtrung:
Oh, I was looking for a similar interactive search feature like `isearch`, but haven't been managed to do that.

I don't use `eaf-file-manager`, but will take a look to see how it handles the search

PR #864, opened 2021-10-23
Interactive search for browser

taquangtrung:
Hi,

I implemented the interactive search for browsers in this PR. The way it works now is quite similar to eaf-file-manager.

Could you take a look and merge if possible?


manateelazycat:
Thanks for patch.

I have push new patch https://github.com/emacs-eaf/emacs-application-framework/commit/d14993ba1b123aca5326f4061c734abc4dba4c1f

You can now press C-s or C-r in minibuffer after type search keyword, try it! ;)

PR #866, opened 2021-10-25
WIP: Installation with straight

ThibautVerron:
Hi,

Here is the current state of the code to check if the package is newly installed/updated, and to do all the copying/building automatically if that is the case.

Current problems:
- the `app` directory and subdirectories don't get added to the load path on the first pass
- the code for melpa is untested (and probably does not work)


MatthewZMD:
After a second thought, I'm thinking can we simply symlink the entire app/ to the build directory? Copy pasting everything is not a very neat solution.

ThibautVerron:
The problem with that is that it would put the built files in the same directory as the source files, breaking straight's intent of separating them.

What we could do instead is what straight does: recursively traverse the directory structure, symlinking all files in the process. It's somewhat tricky, I guess.

An easier option might be to simply move the app directory over to the build directory.

MatthewZMD:
> What we could do instead is what straight does: recursively traverse the directory structure, symlinking all files in the process. It's somewhat tricky, I guess.

Yes, this is what I prefer. Copying everything is easy, but isn't really neat. EAF apps have way too many files, especially the ones in node_modules.

ThibautVerron:
> Yes, this is what I prefer. Copying everything is easy, but isn't really neat. EAF apps have way too many files, especially the ones in node_modules.

Fair enough. I will have a look at what straight does. I remember that one part that's tricky is how windows handles symlinks. Does eaf support windows?

MatthewZMD:
> > Yes, this is what I prefer. Copying everything is easy, but isn't really neat. EAF apps have way too many files, especially the ones in node_modules.
> 
> Fair enough. I will have a look at what straight does.

Thank you, really appreciate your effort :)

> 
> I remember that one part that's tricky is linked to how windows handles symlinks. Does eaf support windows?

Yes, both Windows and macOS are supported by the community.

ThibautVerron:
Actually, I think I'm overthinking this.

There are two targets for the change:
- straight users, with a separate build and source directory
- other users, presumably installing directly via git, and building in the source directory

So we can probably assume that users who want to have the build and source directories separated are using straight, and just call `straight--symlink-recursively`. If the function is not available (but the build and source directories are different), we can fall back to symlinking the whole directory.

MatthewZMD:
> So we can probably assume that users who want to have the build and source directories separated are using straight, and just call `straight--symlink-recursively`. If the function is not available (but the build and source directories are different), we can fall back to symlinking the whole directory.

I had no idea about this function, if it works, that'd be awesome

eaf.el
```diff
@@ -113,14 +115,45 @@ A bookmark handler function is used as
...
+	  (const :tag "EAF Music Player" eaf-music-player)
+	  (const :tag "EAF Org Previewer" eaf-org-previewer)
+	  (const :tag "EAF PDF Viewer" eaf-pdf-viewer)
+	  (const :tag "EAF System Monitor" eaf-system-monitor)
+	  (const :tag "EAF Terminal" eaf-terminal)
+	  (const :tag "EAF Video Player" eaf-video-player)
+	  (const :tag "EAF Vue Demo" eaf-vue-demo)
+	  (const :tag "EAF NetEase Cloud Music" eaf-netease-cloud-music)
+	  (const :tag "EAF RSS Reader" eaf-rss-reader)
+))
```
MatthewZMD, line 146: Please simply read the `applications.json` to retrieve the available applications. The EAF code shouldn't explicitly list the apps.

MatthewZMD:
Also, do let me know when you feel the PR is ready for review. :)

eaf.el
ThibautVerron: I agree it would be much better, but I have no idea how to do it. Do you have experience with json in emacs?

As it stands, the README also repeats the list of apps.

ThibautVerron:
I still need to run more tests on the latest code. 

And the problem with `eaf-add-app-dirs-to-load-path` not adding the path when run after installing the apps is still unsolved. I am completely clueless there.

Sorry for the slow progress!

eaf.el
MatthewZMD: > As it stands, the README also repeats the list of apps.

You're right, we missed that part. :)

> I agree it would be much better, but I have no idea how to do it. Do you have experience with json in emacs?

Parsing json isn't too hard in Emacs, I recommend you to have a look at [this part](https://www.gnu.org/software/emacs/manual/html_node/elisp/Parsing-JSON.html). Let me know if you need some help.

ThibautVerron: Okay, thanks for the pointer. 

I was also scared of making the whole thing work with defcustom, but it turns out that passing the output of a suitable function to the :type keyword works, no need to fiddle with macros.

manateelazycat:
(defun my-eaf-install-deps(app-dir)
  "Install deps from dependencies.json."
  (let* ((deps-dict (with-temp-buffer
                      (insert-file-contents (expand-file-name "dependencies.json" app-dir))
                      (json-parse-string (buffer-string))))
         (pip-deps (gethash "win32" (or (gethash "pip" deps-dict) (make-hash-table))))
         (vue-install (gethash "vue_install" deps-dict))
         (npm-install (gethash "npm_install" deps-dict))
         (npm-rebuild (gethash "npm_rebuild" deps-dict))
         (npm-cmd (if (memq system-type '(cygwin windows-nt ms-dos)) "npm.cmd" "npm")))
    (when pip-deps
      (dolist (pkg (append pip-deps nil))
        (message "%s" (shell-command-to-string (format "pip install %s" pkg)))))
    (when vue-install
      (let ((default-directory app-dir))
        (message "%s" (shell-command-to-string (format "%s install" npm-cmd)))
        (message "%s" (shell-command-to-string (format "%s run build" npm-cmd)))))
    (when npm-install
      (let ((default-directory app-dir))
        (message "%s" (shell-command-to-string (format "%s install" npm-cmd)))))
    (when npm-rebuild
      (let ((default-directory app-dir))
        (message "%s" (shell-command-to-string (format "%s rebuild" npm-cmd)))))))

(use-package eaf
  :straight (eaf :type git :host github :repo "emacs-eaf/emacs-application-framework"
                 :files ("*")
                 :post-build ("python" "install-eaf.py" "--install-core-deps"))
  :config
  (when IS-WINDOWS
    (setq eaf-python-command "D:\\Applications\\Scoop\\apps\\python\\current\\python.exe")
    (setq eaf-wm-name "windows")))

(use-package eaf-demo
  :after (eaf)
  :straight (eaf-demo :type git :host github :repo "emacs-eaf/eaf-demo" :files ("*")))

(use-package eaf-browser
  :after (eaf)
  :straight (eaf-browser :type git :host github :repo "emacs-eaf/eaf-browser" :files ("*")
                          :post-build (my-eaf-install-deps (straight--build-dir "eaf-browser"))))

(use-package eaf-pdf-viewer
  :after (eaf)
  :straight (eaf-pdf-viewer :type git :host github :repo "emacs-eaf/eaf-pdf-viewer" :files ("*")
                            :post-build (my-eaf-install-deps (straight--build-dir "eaf-pdf-viewer"))))

(use-package eaf-markdown-previewer
  :after (eaf)
  :straight (eaf-markdown-previewer :type git :host github :repo "emacs-eaf/eaf-markdown-previewer" :files ("*")
                                    :post-build (my-eaf-install-deps (straight--build-dir "eaf-markdown-previewer"))))

Above is copy from https://emacs-china.org/t/straight-eaf/19929

manateelazycat:
Maybe we should use :post-build of striaght to install EAF dependencies.

ThibautVerron:
Sorry for the delayed reply. I'm not sure if this suggestion is the best solution to the problem. Using post-build to install the packages was always an option, but it doesn't take care of linking/copying them, building the elc or correctly populating the load path.

If I understand correctly, what this patch is doing instead is directly downloading and building the apps in the straight build directory. 

It seems somewhat contrary to straight's philosophy. 
In particular, straight considers that invalidating and rebuilding all packages is an acceptable way of dealing with an unexpected situation: it's a relatively cheap step in general, and it can be done offline.
This post-build on the other hand can be somewhat heavier, and requires internet access. It's not just about eaf, it will block the rebuilding of the entire package database.

I also don't see how this patch deals with the sudo issue for when emacs' initialization is headless (e.g. with --daemon).

manateelazycat:
This PR is no response since 2021-12-07, I don't know this patch whether works since I don't use straight.

I thought this patch was deserted, and someone told me we can use :post-build to fix build works for straight, not need those patch.

So I close this PR, no other meaning.

manateelazycat:
> It's not deserted at all, I'm still monitoring comments, just I have nothing new to report. Sorry, it's a very busy period for me so I didn't have time to evaluate the workaround you posted earlier this week until today.
> 
> As for the current state, as I said back in December and already in October, I'm completely clueless as to why the function `eaf-add-app-dirs-to-load-path` does not do what it claims to do when run for a second time. I don't understand the flow of the function, so I don't think I can debug it, or even understand its limitations, without help.
> 
> It is the last obstacle I see. As it stands, the patch works, only it requires a restart to update the load-path. And also, it was only tested by me, so certainly not enough.

If just one problem that eaf-add-app-dirs-to-load-path not works, you can try remove all paths that prefix match EAF path by ```(delete path load-path)```, then call eaf-add-app-dirs-to-load-path again.

eaf.el
```diff
@@ -1677,15 +1704,50 @@ It currently identifies PDF, videos, images, and mindmap file extensions."
...
+				  eaf-python-command " install-eaf.py" " --install "
+				  (mapconcat 'symbol-name apps " "))
+				 output-buffer)
+	    (get-buffer-process output-buffer))))
+    (if (process-live-p proc)
+	(set-process-sentinel proc #'eaf--post-install-sentinel)
+      (eaf--post-install))))
+
+(defun eaf--post-install-sentinel (process signal)
+  (when (memq (process-status process) '(exit signal))
```
manateelazycat, line 1730: > And the problem with eaf-add-app-dirs-to-load-path not adding the path when run after installing the apps is still unsolved. I am completely clueless there.

Are you sure process-status is `exit` ? 

A process finish, should return `finished` or `exited` status, please check https://www.gnu.org/software/emacs/manual/html_node/elisp/Sentinels.html

ThibautVerron:
> > It's not deserted at all, I'm still monitoring comments, just I have nothing new to report. Sorry, it's a very busy period for me so I didn't have time to evaluate the workaround you posted earlier this week until today.
> > As for the current state, as I said back in December and already in October, I'm completely clueless as to why the function `eaf-add-app-dirs-to-load-path` does not do what it claims to do when run for a second time. I don't understand the flow of the function, so I don't think I can debug it, or even understand its limitations, without help.
> > It is the last obstacle I see. As it stands, the patch works, only it requires a restart to update the load-path. And also, it was only tested by me, so certainly not enough.
> 
> If just one problem that eaf-add-app-dirs-to-load-path not works, you can try remove all paths that prefix match EAF path by `(delete path load-path)`, then call eaf-add-app-dirs-to-load-path again.

Thanks for the hint, and for reopening the issue. I will try that and let you know.

eaf.el
manateelazycat: please reference my another plugin

https://github.com/manateelazycat/multi-term/blob/017c77c550115936860e2ea71b88e585371475d5/multi-term.el#L677

ThibautVerron: Yes, that would also work. Is there an advantage in using that over querying process-status? 

In any case, I am sure that the post-install function is being called.

manateelazycat: Is post-install call after install-eaf.el finish EAF application git clone?

ThibautVerron: The signal is "finished", so it looks like it. Also, the symlink and byte compilation steps of post-install correctly pick both eaf and the newly installed apps.

Only the load path doesn't get updated.

eaf.el
```diff
@@ -1749,7 +1749,7 @@ For a full `install-eaf.py' experience, refer to `--help' and run in a terminal.
       (byte-recompile-directory eaf-build-dir 0))
   (message "Cleaning load path")
   (setq load-path
-	(cl-remove-if (lambda (s) (string-prefix-p eaf-build-dir s)) load-path))
+	(cl-remove-if (lambda (s) (string-prefix-p (directory-file-name eaf-build-dir) s)) load-path))
```
manateelazycat, line 1752: Use let expression move directory-file-name out of loop, will improve performance.

ThibautVerron:
> > > It's not deserted at all, I'm still monitoring comments, just I have nothing new to report. Sorry, it's a very busy period for me so I didn't have time to evaluate the workaround you posted earlier this week until today.
> > > As for the current state, as I said back in December and already in October, I'm completely clueless as to why the function `eaf-add-app-dirs-to-load-path` does not do what it claims to do when run for a second time. I don't understand the flow of the function, so I don't think I can debug it, or even understand its limitations, without help.
> > > It is the last obstacle I see. As it stands, the patch works, only it requires a restart to update the load-path. And also, it was only tested by me, so certainly not enough.
> > 
> > 
> > If just one problem that eaf-add-app-dirs-to-load-path not works, you can try remove all paths that prefix match EAF path by `(delete path load-path)`, then call eaf-add-app-dirs-to-load-path again.
> 
> Thanks for the hint, and for reopening the issue. I will try that and let you know.

After testing, no, it still doesn't pick up the new paths.

(Sorry for the spam with the commits, I need to push a new commit every time I test, and I didn't think of creating a new branch.)

```
ELISP> (eaf-add-app-dirs-to-load-path)
nil
ELISP> load-path
("/home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/" "/home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/all-the-icons" "/home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/straight" "/usr/local/emacs-dev/share/emacs/28.0.50/lisp" ...)

ELISP> (directory-files eaf-build-dir)
("." ".." "app" "core" "eaf-autoloads.el" "eaf-version-file.txt" "eaf.el" "eaf.elc" "eaf.py" "extension" "img" "sync-eaf-resources.py")

ELISP> (directory-files (expand-file-name "app" eaf-build-dir))
("." ".." "demo")

ELISP> (directory-files (expand-file-name "app/demo" eaf-build-dir))
("." ".." ".github" ".gitignore" "LICENSE" "README.md" "buffer.py" "eaf-demo.el" "eaf-demo.elc")
```

manateelazycat:
> > > > It's not deserted at all, I'm still monitoring comments, just I have nothing new to report. Sorry, it's a very busy period for me so I didn't have time to evaluate the workaround you posted earlier this week until today.
> > > > As for the current state, as I said back in December and already in October, I'm completely clueless as to why the function `eaf-add-app-dirs-to-load-path` does not do what it claims to do when run for a second time. I don't understand the flow of the function, so I don't think I can debug it, or even understand its limitations, without help.
> > > > It is the last obstacle I see. As it stands, the patch works, only it requires a restart to update the load-path. And also, it was only tested by me, so certainly not enough.
> > > 
> > > 
> > > If just one problem that eaf-add-app-dirs-to-load-path not works, you can try remove all paths that prefix match EAF path by `(delete path load-path)`, then call eaf-add-app-dirs-to-load-path again.
> > 
> > 
> > Thanks for the hint, and for reopening the issue. I will try that and let you know.
> 
> After testing, no, it still doesn't pick up the new paths.
> 
> (Sorry for the spam with the commits, I need to push a new commit every time I test, and I didn't think of creating a new branch.)
> 
> ```
> ELISP> (eaf-add-app-dirs-to-load-path)
> nil
> ELISP> load-path
> ("/home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/" "/home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/all-the-icons" "/home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/straight" "/usr/local/emacs-dev/share/emacs/28.0.50/lisp" ...)
> 
> ELISP> (directory-files eaf-build-dir)
> ("." ".." "app" "core" "eaf-autoloads.el" "eaf-version-file.txt" "eaf.el" "eaf.elc" "eaf.py" "extension" "img" "sync-eaf-resources.py")
> 
> ELISP> (directory-files (expand-file-name "app" eaf-build-dir))
> ("." ".." "demo")
> 
> ELISP> (directory-files (expand-file-name "app/demo" eaf-build-dir))
> ("." ".." ".github" ".gitignore" "LICENSE" "README.md" "buffer.py" "eaf-demo.el" "eaf-demo.elc")
> ```

I suggest you add a `message` statement in the `eaf-add-app-dirs-to-load-path` function, I doubt that when you call post-install, the emacs-application-framework/apps directory not include any *.el files.

ThibautVerron:
At which point in the function should the message be?

I'm sure that the .el files are there when the post-install function is called, because the byte compiler sees them:

```
Running post install
Symlinking app directory
Byte-compiling
Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/... [2 times]
Compiling /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/eaf.el...done
Wrote /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/eaf.elc
Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/...
Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app...
Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/core... [2 times]
Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/extension... [6 times]
Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/img...
Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo...
Compiling /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo/eaf-demo.el...done
Wrote /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo/eaf-demo.elc
Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo...
Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/core/js...
Done (Total of 2 files compiled, 7 skipped in 4 directories)
Cleaning load path
Updating load path
Done

```

manateelazycat:
> At which point in the function should the message be?
> 
> I'm sure that the .el files are there when the post-install function is called, because the byte compiler sees them:
> 
> ```
> Running post install
> Symlinking app directory
> Byte-compiling
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/... [2 times]
> Compiling /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/eaf.el...done
> Wrote /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/eaf.elc
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/...
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app...
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/core... [2 times]
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/extension... [6 times]
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/img...
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo...
> Compiling /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo/eaf-demo.el...done
> Wrote /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo/eaf-demo.elc
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo...
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/core/js...
> Done (Total of 2 files compiled, 7 skipped in 4 directories)
> Cleaning load path
> Updating load path
> Done
> ```

Print https://github.com/emacs-eaf/emacs-application-framework/blob/dcfa39e5616a2f71b4a8c697d6b0d25bc2bd2b2b/eaf.el#L128 this-dir

manateelazycat:
> At which point in the function should the message be?
> 
> I'm sure that the .el files are there when the post-install function is called, because the byte compiler sees them:
> 
> ```
> Running post install
> Symlinking app directory
> Byte-compiling
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/... [2 times]
> Compiling /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/eaf.el...done
> Wrote /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/eaf.elc
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/...
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app...
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/core... [2 times]
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/extension... [6 times]
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/img...
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo...
> Compiling /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo/eaf-demo.el...done
> Wrote /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo/eaf-demo.elc
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo...
> Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/core/js...
> Done (Total of 2 files compiled, 7 skipped in 4 directories)
> Cleaning load path
> Updating load path
> Done
> ```

Just app/demo/eaf-demo.el found? I haven't seen other apps elisp file byte compile.

eaf.el
manateelazycat: > The signal is "finished", so it looks like it. Also, the symlink and byte compilation steps of post-install both correctly pick eaf and the newly installed apps.
> 
> Only the load path doesn't get updated.

If signal is `finished`, why you use `exit` signal to compare ? 
If signal is mismatch, `eaf--post-install` will call in `eaf--post-install-sentinel `?

ThibautVerron:
It's as expected, my test process only installs eaf-demo: https://github.com/ThibautVerron/emacs-application-framework/blob/installation-test/test.el

ThibautVerron:
> > At which point in the function should the message be?
> > I'm sure that the .el files are there when the post-install function is called, because the byte compiler sees them:
> > ```
> > Running post install
> > Symlinking app directory
> > Byte-compiling
> > Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/... [2 times]
> > Compiling /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/eaf.el...done
> > Wrote /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/eaf.elc
> > Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/...
> > Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app...
> > Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/core... [2 times]
> > Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/extension... [6 times]
> > Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/img...
> > Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo...
> > Compiling /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo/eaf-demo.el...done
> > Wrote /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo/eaf-demo.elc
> > Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/app/demo...
> > Checking /home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/core/js...
> > Done (Total of 2 files compiled, 7 skipped in 4 directories)
> > Cleaning load path
> > Updating load path
> > Done
> > ```
> 
> Print
> 
> https://github.com/emacs-eaf/emacs-application-framework/blob/dcfa39e5616a2f71b4a8c697d6b0d25bc2bd2b2b/eaf.el#L128
> 
> this-dir

With `(message "%s : %s" this-dir contents)`, I get:

```
Updating load path
/home/thibaut/Development/emacs-application-framework/.emacs-eaf.d/build/eaf/: (. .. app core eaf-autoloads.el eaf-version-file.txt eaf.el eaf.elc eaf.py extension img sync-eaf-resources.py)
Done
```

eaf.el
ThibautVerron: Signal is the string `"finished"`, the process status is the symbol `exit`. Another accepted value of the process status for the sentinel would be the symbol `signal` (reached if e.g. one were to issue `killall python3` during the installation process).

Comparing a standard list of symbols for possible statuses seems more robust (and efficient) than matching against a non-exhaustive list of signal strings, no?

ThibautVerron:
It seems that the reason the loop doesn't continue is:

https://github.com/emacs-eaf/emacs-application-framework/blob/dcfa39e5616a2f71b4a8c697d6b0d25bc2bd2b2b/eaf.el#L137

What is this line for?

https://github.com/emacs-eaf/emacs-application-framework/blob/dcfa39e5616a2f71b4a8c697d6b0d25bc2bd2b2b/eaf.el#L138

manateelazycat:
> It seems that the reason the loop doesn't continue is:
> 
> https://github.com/emacs-eaf/emacs-application-framework/blob/dcfa39e5616a2f71b4a8c697d6b0d25bc2bd2b2b/eaf.el#L137
> 
> What is this line for?
> 
> https://github.com/emacs-eaf/emacs-application-framework/blob/dcfa39e5616a2f71b4a8c697d6b0d25bc2bd2b2b/eaf.el#L138

File-Node ID, avoid add duplicate path in load-path.

ThibautVerron:
Aha okay, then there's the bug. Currently the function can't process subdirectories of a directory listed in that list.

I see different ways out, but the easiest is probably to reset this variable between calls to the function, and call `cl-remove-duplicates` after the second run (or at the end of `add-subdirs...`). We won't have too many duplicates, since the variable will still do its job within a function call.
The cost of removing duplicates is, but even on my relatively bloated emacs config, `load-path` is less than 200 elements. A one-time cost of 40000 string comparisons should be acceptable...

ThibautVerron:
With this change it all seems to work. 

Thanks a lot for the help!

manateelazycat:
> With this change it all seems to work.
> 
> Thanks a lot for the help!

I have push new commit https://github.com/emacs-eaf/emacs-application-framework/commit/bba64c7755eda0d81f3b9fb7d4996ed83dc9420c that I re-implement `eaf-add-subdirs-to-load-path` .

I believe `eaf-add-subdirs-to-load-path` can work, can you try again?

Thanks for effort! ;)

manateelazycat:
@ThibautVerron I have merge this PR, but remove auto-install code and make byte-compile option default as nil.

Thanks for effort, please send new PR if still have problem with Straight. ;)

ThibautVerron:
> > > @ThibautVerron I have merge this PR, but remove auto-install code
> > 
> > 
> > Can I ask why?
> > > and make byte-compile option default as nil.
> > 
> > 
> > The default behavior in emacs (both the built-in package.el and straight.el) is to byte compile the elisp files. I'm not even sure if there is a way to disable this in package.el.
> > Adding a variable for it is harmless, but I don't see any reason to have it disabled by default.
> > Sorry, I didn't have time to test the new function, but if it's merged already I guess it will be tested one way or another. :)
> 
> Because we should not call eaf-install-and-update when user first time load eaf.el, it's not right.

We can make the default value of auto-install nil. But why force the user to run the install function manually, if it can be done automatically?

> 
> And user can use any way to install EAF, EAF needless to follow the default behavior of package.el or straight, if user wan't to byte-compile, he can turn on in config.

But if byte compiling is the default behavior (and not even customizable) everywhere else, why would a user even look into the config to see that eaf needs to pull elisp on its own, and by default doesn't byte compile them?

There is zero harm in byte-compiling. Why force gazillions of users to change the variable?

manateelazycat:
> > > > @ThibautVerron I have merge this PR, but remove auto-install code
> > > 
> > > 
> > > Can I ask why?
> > > > and make byte-compile option default as nil.
> > > 
> > > 
> > > The default behavior in emacs (both the built-in package.el and straight.el) is to byte compile the elisp files. I'm not even sure if there is a way to disable this in package.el.
> > > Adding a variable for it is harmless, but I don't see any reason to have it disabled by default.
> > > Sorry, I didn't have time to test the new function, but if it's merged already I guess it will be tested one way or another. :)
> > 
> > 
> > Because we should not call eaf-install-and-update when user first time load eaf.el, it's not right.
> 
> We can make the default value of auto-install nil. But why force the user to run the install function manually, if it can be done automatically?
> 
> > And user can use any way to install EAF, EAF needless to follow the default behavior of package.el or straight, if user wan't to byte-compile, he can turn on in config.
> 
> But if byte compiling is the default behavior (and not even customizable) everywhere else, why would a user even look into the config to see that eaf needs to pull elisp on its own, and by default doesn't byte compile them?
> 
> There is zero harm in byte-compiling. Why force gazillions of users to change the variable?
> 
> Original discussion for reference (but without much more content): [#799 (comment)](https://github.com/emacs-eaf/emacs-application-framework/discussions/799#discussioncomment-1322869)

Auto-Install is not right policy, and many countries (such as China) can't access github directly, we need use proxy to install, if you force auto-install, user will failed.

If you force byte-compile el file, elc only can work in current OS (such as Linux), when user switch to other OS (such as macOS), elc can't work, need user byte-compile all el file manually, it's important that don't byte-compile if user want "one config for multiple operating system", so byte-compile is not zero harm.

The world is not the same, we should give users more choices, we way is not only way.

PR #876, opened 2021-11-21
fix bug in duplicating page

taquangtrung:
Hi,

The URL of the current page to be duplicated should be `self.url`, but not `current_url`.
`current_url` isn't updated anywhere in the code.


manateelazycat:
thanks for

PR #894, opened 2021-12-20
Toggle Qt.WindowStaysOnTopHint on focus change instead of using the *eaf-temp* buffer

### LINKED ISSUE #835: macOS: EAF buffers always stay on top of other applications
Hi,

**Describe the bug**

When I'm opening an EAF webpage and then press Alt-Tab to switch to another application, the EAF webpage always stays on top of the new application.

**To Reproduce**

- Open Google Chrome
- Open Emacs and run `eaf-search-it` to open a webpage.
- Press Alt-Tab to switch to Google Chrome, the EAF webpage always stays on top of Google Chrome.

Screenshot 2021-08-27 at 9 04 33 AM

**Expected behavior**

After switching to Google Chrome, the EAF webpage should stay below it.

**Versions (please complete the following info):**
 - Distro and DE/WM: macOS BigSur 11.3
 - Versions of Dependencies: eaf-20210826.100
 - M-x emacs-version: 27.2

Gira-X:
Tagging @taquangtrung.

Maybe @lhpfvs can also take a look as the initial code with the temporary buffer was merged in https://github.com/emacs-eaf/emacs-application-framework/pull/616.
But I am not sure if the current code works with multiple frames as I don't need them on macOS.

Resolve #835

### DIFF
--- a/core/view.py
+++ b/core/view.py
@@ -145,6 +145,14 @@ class View(QWidget):
 
         qwindow.setPosition(QPoint(self.x, self.y))
 
+    def handle_emacs_focus_in(self):
+        self.setWindowFlag(Qt.WindowStaysOnTopHint, True)
+        self.show()
+
+    def handle_emacs_focus_out(self):
+        self.setWindowFlag(Qt.WindowStaysOnTopHint, False)
+        self.hide()
+
     def destroy_view(self):
         # print("Destroy: ", self.buffer.url)
         self.destroy()
--- a/eaf.el
+++ b/eaf.el
@@ -364,11 +364,12 @@ been initialized."
     "Execute EAF Python process under Rosetta2"
     :type 'boolean)
 
-  (defvar eaf--mac-switch-to-python nil
-    "Record if Emacs switchs to Python process")
+  (defvar eaf--mac-focus-change-timer nil
+    "Use timer to ignore spurious focus events.
 
-  (defvar eaf--mac-has-focus t
-    "Record if Emacs has focus"))
+See
+https://old.reddit.com/r/emacs/comments/\
+kxsgtn/ignore_spurious_focus_events_for/"))
 
 (defcustom eaf-name "*eaf*"
   "Name of EAF buffer."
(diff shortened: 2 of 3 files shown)


manateelazycat:
Thanks for patch, awesome man. ;)

taquangtrung:
@Gira-X: I just tested your patch. Now Alt-tab can work nicely and EAF buffers will not stay on top of other apps when Emacs is put hidden 👍!

On the other hand, I notice that when Emacs is open, if I click mouse directly on an EAF buffer, it may disappear. If I click mouse on the same place again, then it re-appears. Not sure if this also happens to you?

Gira-X:
Ah nice catch with the mouse clicking. I don't really use the mouse in Emacs so I never tested this.

Clicking on an EAF buffer triggers `after-focus-change-function` twice. First with `nil` and then instantly with `t` again which is indeed annoying and should be fixed. This is regardless even if Emacs was focused before.

I think this happens because the Emacs application loses focus for a split second to activate the Python window.

I'll investigate a fix, but I have the feeling that I might have to bring back the `app-frontmost --name` shell call which caused the slight and annoying Alt-Tab delay in the first place.
Maybe I'll guard it with a `defcustom` variable though, so users like me can benefit from the quick Alt-Tab.

taquangtrung:
@Gira-X: thanks for the update!

The other issue that I encountered is that when putting Emacs in the full screen mode (F11), all EAF buffers will disappear. You can see in my screenshots below. 

**Normal view:**

Screenshot 2021-12-21 at 2 39 41 AM

**Full-screen mode:**

Screenshot 2021-12-21 at 2 39 49 AM

Does it happens to you too?

I'm not sure whether it's due to my Emacs distribution: I'm using Emacs macport version.

Gira-X:
> I'm using the Emacs Mac port version

I am also using it and I can also reproduce this issue. But I also never noticed it because I run Emacs with `mac-use-title-bar` set to nil so the native macOS title bar is never displayed for me. Oh well...

---

And thanks for reporting the bugs. Looks like there is still some work ahead :)

@manateelazycat I suggest you revert this merge and we instead cache the result of `(shell-command-to-string "app-frontmost --name")` in `eaf--mac-focus-change`. There is no need to call it twice as it is slow and will lead to an already faster application switch.

manateelazycat:
> > I'm using the Emacs Mac port version
> 
> I am also using it and I can also reproduce this issue. But I also never noticed it because I run Emacs with `mac-use-title-bar` set to nil so the native macOS title bar is never displayed for me. Oh well...
> 
> And thanks for reporting the bugs. Looks like there is still some work ahead :)
> 
> @manateelazycat I suggest you revert this merge and we instead cache the result of `(shell-command-to-string "app-frontmost --name")` in `eaf--mac-focus-change`. There is no need to call it twice as it is slow and will lead to an already faster application switch.

please send new fix patch, I have did many refactory works after your commit, thanks

Gira-X:
> please send new fix patch, I have did many refactory work after your commit, thanks

Check out https://github.com/emacs-eaf/emacs-application-framework/pull/898

PR #898, opened 2021-12-21
Add eaf--mac-safe-focus-change to toggle between unsafe and safe focus change behavior

Gira-X:
This MR implements the `defcustom` `eaf--mac-safe-focus-change` which is set to `t` by default and which implements the behavior before https://github.com/emacs-eaf/emacs-application-framework/pull/894 was merged. The only change is that the `app-frontmost --name` shell command is cached and not called twice as before which improves application switch performance.

Setting it to `nil` implements the behavior which was implemented in https://github.com/emacs-eaf/emacs-application-framework/pull/894

This variable can be toggled on the fly.

@taquangtrung, it would be great if you could test this as well before merging this :)

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -988,15 +988,15 @@ kxsgtn/ignore_spurious_focus_events_for/")
       (set-window-configuration
        (frame-parameter (selected-frame) 'eaf--mac-frame)))
 
-    (defun eaf--mac-unsafe-focus-out ()
+    (defun eaf--mac-unsafe-focus-out (&optional frame)
       (eaf-call-async "mac_handle_emacs_focus_out")
-      (set-frame-parameter (selected-frame) 'eaf--mac-frame
+      (set-frame-parameter (or frame (selected-frame)) 'eaf--mac-frame
                            (current-window-configuration)))
 
-    (defun eaf--mac-delete-frame-handler ()
+    (defun eaf--mac-delete-frame-handler (frame)
       (if eaf--mac-safe-focus-change
-          (eaf--mac-focus-out)
-        (eaf--mac-unsafe-focus-out)))
+          (eaf--mac-focus-out frame)
+        (eaf--mac-unsafe-focus-out frame)))
 
     (add-function :after after-focus-change-function #'eaf--mac-focus-change)
     (add-to-list 'delete-frame-functions #'eaf--mac-delete-frame-handler)))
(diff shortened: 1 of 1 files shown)


manateelazycat:
Thanks for patch, send new patch if have futher problem.

taquangtrung:
@Gira-X: I tested this new patch and it works well! Well done!

FYI, I also created this function to quickly toggle the state of `eaf--stay-on-top-safe-focus-change`:

```emacs-lisp
(defun eaf-mac-toggle-stay-on-top-safe-focus-change ()
  "Toggle stay-on-top focus change for macOS."
  (interactive)
  (setq eaf--stay-on-top-safe-focus-change
        (not eaf--stay-on-top-safe-focus-change))
  (if eaf--stay-on-top-safe-focus-change
      (message "EAF: enabled stay-on-top focus change!")
    (message "EAF: disabled stay-on-top focus change!")))
```

manateelazycat:
@Gira-X Can you test ```self.setWindowFlags(Qt.FramelessWindowHint | Qt.BypassWindowManagerHint | Qt.NoDropShadowWindowHint)``` window flag on macOS platform?

Flag Qt.BypassWindowManagerHint mean EAF python icon won't show in taskbar, even EAF python not switch by Alt + Tab.

If flag Qt.BypassWindowManagerHint works on macOS, then you don't need use app-frontmost to test active process is EAF or Emacs.

Gira-X:
> Flag Qt.BypassWindowManagerHint mean EAF python icon won't show in taskbar, even EAF python not switch by Alt + Tab.

I tried out `self.setWindowFlags(Qt.FramelessWindowHint | Qt.BypassWindowManagerHint | Qt.NoDropShadowWindowHint)` in `view.py` but on macOS 10.14.6 there is still the `Python` Dock application.

I used it with `eaf--stay-on-top-safe-focus-change` set to `nil` and I experience some issues when this flag is enabled, for instance when deleting other Emacs windows when I am inside an EAF buffer, the EAF buffer is blank afterwards, so I would not recommend using it.

manateelazycat:
> > Flag Qt.BypassWindowManagerHint mean EAF python icon won't show in taskbar, even EAF python not switch by Alt + Tab.
> 
> I tried out `self.setWindowFlags(Qt.FramelessWindowHint | Qt.BypassWindowManagerHint | Qt.NoDropShadowWindowHint)` in `view.py` but on macOS 10.14.6 there is still the `Python` Dock application.
> 
> I used it with `eaf--stay-on-top-safe-focus-change` set to `nil` and I experience some issues when this flag is enabled, for instance when deleting other Emacs windows when I am inside an EAF buffer, the EAF buffer is blank afterwards, so I would not recommend using it.

thanks for help

PR #899, opened 2021-12-21
Add an error message when the Emacs frame ID is nil

roygbyte:
It appears to me that EAF doesn't work when Emacs is launched from a shell. Is this a known limitation or an error in implementation? Or, is it a problem with my machine? I am running Emacs 26.3 on Ubuntu with bash version 5.0.17. When I launch Emacs from a shell, EAF functions appear and I can launch the demo `M-x eaf-open-demo`. But: I get the following message in `*eaf*`:

``` sh
Traceback (most recent call last):                                                                                                                                                                                                            
  File "/home/.../.emacs.d/site-lisp/emacs-application-framework/core/utils.py", line 61, in on_signal_received                                                                                                                          
    self._func(obj, *args, **kwargs)                                                                                                                                                                                                          
  File "/home/.../.emacs.d/site-lisp/emacs-application-framework/eaf.py", line 238, in update_views                                                                                                                                      
    view = View(self.buffer_dict[buffer_id], view_info)                                                                                                                                                                                       
  File "/home/.../.emacs.d/site-lisp/emacs-application-framework/core/view.py", line 52, in __init__                                                                                                                                     
    self.emacs_xid = int(self.emacs_xid)                                                                                                                                                                                                      
ValueError: invalid literal for int() with base 10: 'nil'       
```
However, when I launch Emacs from GUI and run the demo, I see the success message ("Hello, EAF hacker, it's working!!!").  

Anyways, I'm not sure if this is a command problem/issue/confusion. But in this PR I've added an error message that will tell the user that they should try launching Emacs via the GUI if an Emacs frame can't be found. For me, that fixed the problem. I've also added a note to `README` to suggest ways a user can diagnose problems they are having with EAF.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,11 +127,13 @@ You can use below code to load applications `browser` and `pdf-viewer` that you 
 
 #### 5. Hooray!
 
-Congratulations, you just installed EAF! You can try `M-x eaf-open-demo` (that is if you have `demo` installed, of course) to see if everything works properly, and enjoy the new possibilities of Emacs.
+Congratulations, you just installed EAF! You can try `M-x eaf-open-demo` (that is if you have `demo` installed, of course) to see if everything works properly. If the demo is installed but isn't launching, check the EAF buffer (`*eaf*`) to begin diagnosing the problem with your installation. 
 
 I **highly** encourage you to read the [Wiki](#Wiki) and [FAQ](#FAQ) if you have any questions.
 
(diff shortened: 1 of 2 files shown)


manateelazycat:
thanks for patch, but patch position is not correctly, i will add warning when i wake up tomorrow, it's 00:00 in china

manateelazycat:
I have push new commit https://github.com/emacs-eaf/emacs-application-framework/commit/775569bb356784122ff9bcf5c176ba1fa2a6cff2 to fix emacs-xid error 

EAF should work with console mode, I can't make eaf-get-window-allocation return pixel size under console mode now.

If we can got pixel size of window when Emacs running in console, EAF can running in console.

PR #901, opened 2021-12-22
Add darkreader class to markers to not be affected by dark mode

Gira-X:
Right now, the dark mode affects the markers as well.

This is with dark mode disabled:

image

And this is with dark mode enabled where the markers are also affected:

image

---

Adding the `darkreader` class ignores the markers from being affected by the dark theme.

See https://github.com/darkreader/darkreader/issues/4056#issuecomment-723554928


MatthewZMD:
Awesome, thanks for the patch!

PR #903, opened 2021-12-23
Add caret_enable_mark and patch _caret_at_line for initial selection

Gira-X:
Now the initial highlighting with the mark works. And calling `caret_at_line` now works when the mark was already set to select another region.

In some cases, the initial highlighting is still not there. I am not sure why exactly, but this this MR provides an improvement from the current behavior.

Resolve https://github.com/emacs-eaf/eaf-browser/issues/28

### DIFF
--- a/core/webengine.py
+++ b/core/webengine.py
@@ -614,11 +614,16 @@ class BrowserView(QWebEngineView):
         self.eval_js("Marker.gotoMarker('%s', (e) => window.getSelection().collapse(e, 0))" % str(marker))
         self.cleanup_links_dom()
 
-        self.eval_js("CaretBrowsing.setInitialCursor(true);")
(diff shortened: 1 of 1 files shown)


manateelazycat:
I'm not use caret mode, If you find highlight not work, maybe cause by ```QTimer().singleShot(0, lambda : self.triggerPageAction(self.web_page.Unselect))``` in webengine.py

Intro by patch b51c2ee1

Gira-X:
> If you find highlight not work, maybe cause by QTimer().singleShot(0, lambda : self.triggerPageAction(self.web_page.Unselect)) in webengine.py

It's not related. I checked if `clean_search_and_select`  is called but it is not related to `caret_at_line`.

PR #905, opened 2021-12-28
add gentoo distro emerge support

lengyueyang:
Add gentoo distro support, add emerge to dependencies.json and install-eaf.py

### DIFF
--- a/dependencies.json
+++ b/dependencies.json
@@ -17,6 +17,8 @@
         "python3-pyqt5.qtwebengine",
         "python3-pygit2"
     ],
+    "emerge": [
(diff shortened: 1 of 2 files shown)


manateelazycat:
Thanks for patch, I think we need add gentoo packages in depend.json file, not just emacs-application-framework repository, also include all emacs-eaf applications' depend.json.

manateelazycat:
We need add gentoo package in this patch, I regret that there is no progress in this long time, temporarily closed, welcome to PR at any time. ;)

lengyueyang:
OK

PR #912, opened 2022-01-23
Use a list of helpers instead of yay

Ygenks:
Hello. 
My rationale was driver by an error of missing "yay" package manager on the system.
I personally use pacaur, someone else could use old-style yaourt.


MatthewZMD:
Very reasonable, thanks for the patch!

PR #929, opened 2022-03-18
doc: touch up README.

yatsky:
Hi,
This is just a couple of small modifications to the doc to make it look nicer.
Thanks for the great package!


manateelazycat:
thanks for patch, thank you

PR #930, opened 2022-03-19
fix: drag in eaf window when using python3

yuchen-lea:
fix issue here https://emacs-china.org/t/eaf/20252/12


manateelazycat:
Thanks for help

PR #934, opened 2022-03-21
Fix Printing to File on Arch

ragreenburg:
Basically when Arch doesn't have a helper this doesn't right to a file, it errors out saying it doesn't know what "std" is and asking if you mean "str". This fixes that so it writes to std.err


MatthewZMD:
This should be fixed in the latest commit.

PR #935, opened 2022-03-21
Some problems with installing eaf on archlinux

kuxiade:


MatthewZMD:
Thanks for the patch!

PR #946, opened 2022-04-08
bug fix: `eaf-install-and-update`

vritser:
The `--install` flag requires at least one argument, but `eaf-apps-to-install` defaults to nil.

### DIFF
--- a/eaf.el
+++ b/eaf.el
@@ -1697,7 +1697,8 @@ For a full `install-eaf.py' experience, refer to `--help' and run in a terminal.
          (proc
           (progn
             (async-shell-command (concat
-                                  eaf-python-command " install-eaf.py" " --install "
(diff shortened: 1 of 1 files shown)


eaf.el
```diff
@@ -1697,7 +1697,8 @@ For a full `install-eaf.py' experience, refer to `--help' and run in a terminal.
          (proc
           (progn
             (async-shell-command (concat
-                                  eaf-python-command " install-eaf.py" " --install "
+                                  eaf-python-command " install-eaf.py"
+                                  (when (length> apps 0) " --install ")
```
MatthewZMD, line 1701: There is a typo here at `length>`

vritser: Oh, this is a new function in version 28, it should be compatible with the old version, my fault

PR #953, opened 2022-04-16
remove dependency on `eaf-browser`

vritser:
These variables are defined in `eaf-browser`, but only used in the `eaf` project, causing `eaf` to depends on `eaf-browser`, so I moved these variables to `eaf` and removed these of `eaf-browser`.

PR is here: https://github.com/emacs-eaf/eaf-browser/pull/32

### DIFF
--- a/core/webengine.py
+++ b/core/webengine.py
@@ -85,11 +85,11 @@ class BrowserView(QWebEngineView):
         (self.default_zoom, self.show_hover_link,
          self.marker_letters, self.marker_fontsize,
          self.scroll_step) = get_emacs_vars(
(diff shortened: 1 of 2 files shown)


manateelazycat:
Thanks for patch, we should use defalias keep compatible for eaf-browser-* variables, otherwise all EAF user will broken after merge PR.

MatthewZMD:
> Thanks for patch, we should use defalias keep compatible for eaf-browser-* variables, otherwise all EAF user will broken after merge PR.

I think it is the right step to use the more suitable name eaf-webengine, we can tell them to update their variable names in the Discussions page

PR #1002, opened 2022-08-31
Update README.md

mikegilchrist:
Added `(require ...)` commands to `use-package` example code to avoid `Error (use-package) eaf/:config: Symbol's value as variable is void: eaf-pdf-viewer...`


MatthewZMD:
> Please add those in README.zh-CN.md，thanks

It's okay, I've done it for him, he doesn't know Chinese.


Thanks for the patch!

PR #1003, opened 2022-09-09
My 'fixes' were mistakes

mikegilchrist:
While I thought I checked my changes to ensure they worked, it turns out they only work if your init.el has already been evaluated (i.e. if you evaluate the init.el buffer within an emacs session that already tried to load it.).  If you start a new session, these changes lead to an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,10 @@ From here on, you can add the full path to the EAF installation directory to you
       (eaf-browser-continue-where-left-off t)
       (eaf-browser-enable-adblocker t)
       (browse-url-browser-function 'eaf-open-browser)
+      :init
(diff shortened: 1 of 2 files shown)


manateelazycat:
Why this patch remove password type？

If remove password type, some input field won't work.

mikegilchrist:
I'm still learning the ropes for git, but I don't believe I removed the password support.  I tried to revert back to that commit (ea6c0e2).

MatthewZMD:
See files changed for your changes in this PR: https://github.com/emacs-eaf/emacs-application-framework/pull/1003/files

mikegilchrist:
Okay.  My apologies.  I will try again...

PR #1015, opened 2022-09-26
Fix: install script can run on virtual env

milanglacier:
see the discussion here.
https://stackoverflow.com/questions/1871549/determine-if-python-is-running-inside-virtualenv/40099080#40099080


manateelazycat:
thanks

PR #1017, opened 2022-10-05
make zoom step smaller and customizable

taquangtrung:
Hi,

This is a simple PR that allows users to customize the zoom step for eaf browsers.

The default zoom step is 0.25, which is quite large, so I make it an option and set the default value to 0.1 (similar to Google Chrome)

Hope that it can be useful for EAF.


MatthewZMD:
Good patch, thank you

PR #1067, opened 2023-02-27
pass --break-system-packages to pip for EXTERNALLY-MANAGED installations

longminwang:
For pip 23.0.1, a flag --break-system-packages should be passed to pip install --user to install a package on an EXTERNALLY-MANAGED python system. Please see [Implement --break-system-packages for EXTERNALLY-MANAGED installations](https://github.com/pypa/pip/pull/11780) and [PEP 668 – Marking Python base environments as “externally managed”](https://peps.python.org/pep-0668/#specification) for description of EXTERNALLY-MANAGED system.

### DIFF
--- a/install-eaf.py
+++ b/install-eaf.py
@@ -7,6 +7,7 @@ import json
 import os
 import subprocess
 import sys
+import sysconfig
 
 parser = argparse.ArgumentParser()
 parser.add_argument("--install-all-apps", action="store_true",
@@ -142,7 +143,11 @@ def install_sys_deps(distro: str, deps_list):
 
 def install_py_deps(deps_list):
     if sys.prefix == sys.base_prefix:
-        command = [PIP_CMD, 'install', '--user', '-U']
+        # pass --break-system-packages to permit installing packages into EXTERNALLY-MANAGED Python installations. see https://github.com/pypa/pip/issues/11780
+        if os.path.exists(os.path.join(sysconfig.get_path("stdlib", sysconfig.get_default_scheme()),"EXTERNALLY-MANAGED")):
+            command = [PIP_CMD, 'install', '--user', '--break-system-packages', '-U']
+        else:
+            command = [PIP_CMD, 'install', '--user', '-U']
     else:
         # if running on a virtual env, --user option is not valid.
         command = [PIP_CMD, 'install', '-U']


manateelazycat:
Thanks for patch.

What problem is solve by this patch?

manateelazycat:
If add this option, we can break OS's file system?

longminwang:
> Thanks for patch.
> 
> What problem is solve by this patch?

> If add this option, we can break OS's file system?

It behaves the same as `pip3 install --user` with version before 23.0.0, and does not affect the OS's file system. 

Without the flag `--break-system-packages`, on externally-managed-environment, `pip3 install --user` will show the following messages and quit. 

```
❯ pip3 --version
pip 23.0.1 from /usr/lib/python3.10/site-packages/pip (python 3.10)
❯ pip3 install --user PyQt6
error: externally-managed-environment

× This environment is externally managed
╰─> 
    The system-wide Python installation in Gentoo should be maintained
    using the system package manager (e.g. emerge).`
    
    If the package in question is not packaged for Gentoo, please
    consider installing it inside a virtual environment, e.g.:
    
    python -m venv /path/to/venv
    . /path/to/venv/bin/activate
    pip install mypackage
    
    To exit the virtual environment, run:
    
    deactivate
    
    The virtual environment is not deleted, and can be re-entered by
    re-sourcing the activate file.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.

❯ pip3 install --user --break-system-packages PyQt6
Collecting PyQt6
  Using cached PyQt6-6.4.2-cp37-abi3-manylinux_2_28_x86_64.whl (8.0 MB)
Collecting PyQt6-Qt6>=6.4.0
  Using cached PyQt6_Qt6-6.4.2-py3-none-manylinux_2_28_x86_64.whl (56.7 MB)
Requirement already satisfied: PyQt6-sip<14,>=13.4 in /home/sde/.local/lib/python3.10/site-packages (from PyQt6) (13.4.1)
Installing collected packages: PyQt6-Qt6, PyQt6
  Attempting uninstall: PyQt6-Qt6
    Found existing installation: PyQt6-Qt6 6.3.1
    Uninstalling PyQt6-Qt6-6.3.1:
      Successfully uninstalled PyQt6-Qt6-6.3.1
Successfully installed PyQt6-6.4.2 PyQt6-Qt6-6.4.2
```

longminwang:
From the [Specification](https://peps.python.org/pep-0668/#specification) of PEP 668, if `sys.prefix == sys.base_prefix` && a file named EXTERNALLY-MANAGED is present in the system lib direction of python, then the system will be EXTERNALLY-MANAGED. This behavior was introduced in pip version 23.0.0, and the option `--break-system-packages` was introduced in pip 23.0.1.

manateelazycat:
I see, thanks for patch

PR #1099, opened 2023-06-11
Fix `eaf-mail` for new versions of `mu4e`

abougouffa:
Fix `eaf-mu4e-get-html`. New versions of `mu4e` renamed `mu4e~view-message` to `mu4e--view-message`. This PR fixes the issue for new versions while supporting the old ones.


manateelazycat:
thanks

PR #1107, opened 2023-06-30
fix module 'sysconfig' has no attribute 'get_default_scheme'

qlq3763:
on my system using python 3.9, sysconfig has _get_default_scheme instead of get_default_scheme, so just try both


manateelazycat:
Thanks

PR #1127, opened 2023-11-22
Disable reinput completely for Hyprland with Xwayland

mumu-lhl:


manateelazycat:
难道不是 current_desktop in ["sway", "Hyprland"] ？

PR #1144, opened 2024-03-13
Improve input focus and retry mechanism

albert748:
This commit enhances the input focus functionality by extending support to include textarea elements alongside input elements in focus_input.js. It also introduces a retry mechanism in webengine.py for instances where the initial focus attempt fails, ensuring a more reliable focus on input elements.


manateelazycat:
@albert748 Please rebase this patch, thanks

PR #1152, opened 2024-05-29
fixed nix dependency for wmctrl

AkashicLibrary:
at least for me wmctrl fails to fetch with nix profiles install without nixpkgs# in front of it. But I did not test if this is just because I have nix command and flakes enabled.


manateelazycat:
thanks

PR #1153, opened 2024-05-30
fixed nix dependency for nodejs

AkashicLibrary:
just realized nodejs had the same problem as wmctrl and thought I should fix it.
I don't know why this didn't make a problem while installing it on nixos like wmctrl did but I think it is better to change it nonetheless.
I just tested if nix profile install node worked and, surprise it didn't but nix profile install nixpkgs#nodejs seems to work fine. At least when manually doing it.

### DIFF
--- a/dependencies.json
+++ b/dependencies.json
@@ -16,7 +16,7 @@
   ],
   "nix": [
     "nixpkgs#wmctrl",
-    "node"
+    "nixpkgs#nodejs"
   ],
   "emerge": [
     "x11-misc/wmctrl",


AkashicLibrary:
just realized I need to look into it a bit more. Cause it seems to not even try install but maybe it sees nodejs already installed through systempackages but wmctrl not. Found this out after using the install script and checking nix profile list and it was not there (I removed it manually for my tests) so I'm going to try if it installs when I don't have nodejs already installed in systempackages and reply again.

AkashicLibrary:
just checked and it seems it detects nodejs but not wmctrl if installed through systempackages.

manateelazycat:
this patch need merge?

AkashicLibrary:
then it would be needed

manateelazycat:
thanks

PR #1157, opened 2024-10-05
Make it no query to kill processes when quit emacs

thattemperature:
EAF has two processes that need query to kill when quit emacs, but holo-layer doesn't have. I think no needs of query is right. So I compare the code between them and modify the EAF.


manateelazycat:
thanks

Issue #1169, opened 2025-06-12
Wrong respose of mouse forward/back event


thattemperature:
**Describe the bug**

In eaf-browser, when I press the mouse-forward button, it goes forward; and when I release the mouse-forward button, it goes forward again. Same as mouse-back button.

I think the bug is at the `eventFilter` method in webengine.py file, and I am trying to debug.

**Expected behavior**
It should go forward/back only when press button, not on release event.

**Versions (please complete the following info):**
 - Distro and DE/WM: NixOS, Gnome, Wayland
 - Versions of Dependencies: Python: 3.12, PyQt6: 6.9.0, PyQt6-WebEngine: 6.8.0
 - M-x emacs-version: 30.1

manateelazycat:
PR are welcome

PR #1170, opened 2025-06-12
Fix issue #1169.

thattemperature:
I catch the event with "mouse-release" type and "forward/back" button at the end of the `eventFilter` method, do nothing and return true. This can fix the issue #1169, but I am not sure it is the best practice.


manateelazycat:
Thanks

Issue #1040, opened 2022-12-08
Compatability with header-line


JordanAnthonyKing:
**Describe the bug**
I use a headerline, EAF browser renders on top of it (you can just about see it poking out at the top right)
![image](https://user-images.githubusercontent.com/58787893/206573372-0569b2fc-d0ef-42d1-96f2-efe6bb8fd3c9.png)

Also notice how the EAF window doesn't quite fill the window. Is there a way to fix this?

**To Reproduce**
Enable a headerline and start eaf-browser.

**Expected behavior**
Render below the headerline

**Versions (please complete the following info):**
 - Distro and DE/WM: Custom
 - Versions of Dependencies: Latest
 - M-x emacs-version: 29-Master

manateelazycat:
Any PR are welcome

JordanAnthonyKing:
> What is headerline?

https://www.gnu.org/software/emacs/manual/html_node/elisp/Header-Lines.html

manateelazycat:
It's not just header line, your header line is a new plugin that EAF can't works with.

Any PR are welcome.

MatthewZMD:
Yes, my question is, what is your headerline plugin??

JordanAnthonyKing:
None, just a manually created headerline

PR #1178, opened 2025-08-11
Fix issue #1040.

thattemperature:
I think this issue was introduced by the commit b3de6d38e395867895c1030efbdd46f75f2a1f58.

In fact I can not understand that commit. The header line still works well after the tab line was introduced in Emacs 27.0 version. Why it put `window-header-line-height` and `window-tab-line-height` in a mutually exclusive position?


manateelazycat:
Thanks

PR #1179, opened 2025-08-12
Fix "add-to-list" compilation error.

thattemperature:
Compilation error:
`error   ‘add-to-list’ can’t use lexical var ‘environments’; use ‘push’ or ‘cl-pushnew’`

In `add-to-list` document string:
This is meant to be used for adding elements to configuration
variables, such as adding a directory to a path variable
like ‘load-path’, but please do not abuse it to construct
arbitrary lists in Elisp code, where using ‘push’ or ‘cl-pushnew’
will get you more efficient code.


manateelazycat:
Thanks

PR #1181, opened 2025-08-28
Replace the `app-frontmost` dependency with native AppleScript.

witcxc:
#1182 

- Use `osascript -e 'tell application "System Events" to get name of first application process whose frontmost is true'` to detect the currently active application.
- Support runs natively from Mac OS X 10.3（2003） up to latest macOS on both Intel and Apple Silicon, with no additional software required


manateelazycat:
thanks

PR #1188, opened 2026-03-06
Raise exception when command not found and when ensure_pass is True

cinconueves:
If the command is not found, the exception is not raised, but instead returned, which leads to the exception not being logged and the installation silently failing; the end result will be a bad/partial installation yet the user will receive no information about it until they try and run an eaf application.


manateelazycat:
thanks

